PR #9, opened 2024-02-22
Bugfix: fix compiler linker can't find std::filesystem

JackeyLove1:
In some compilers that support only some C++17 features like gcc6.3.0, it may encounter a situation where std filesystem lib  can't be found, like: 
```
/usr/bin/ld: CMakeFiles/gemma.dir/gemma.cc.o: in function `std::filesystem::__cxx11::path::path<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::filesystem::__cxx11::path>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::filesystem::__cxx11::path::format)':
[gemma.cc](http://gemma.cc/):(.text._ZNSt10filesystem7__cxx114pathC2INSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEES1_EERKT_NS1_6formatE[_ZNSt10filesystem7__cxx114pathC5INSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEES1_EERKT_NS1_6formatE]+0x74): undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
```
So we need to explicitly link the std::filesystem library in cmake.


jan-wassenberg:
Thanks! Does this also work for clang, or do we require some logic to detect which one to link against? Something like https://github.com/ned14/llfio/issues/52 ?

CMakeLists.txt
```diff
@@ -64,7 +64,7 @@ endif()
 add_executable(gemma run.cc)
 target_sources(gemma PRIVATE ${SOURCES})
 set_property(TARGET gemma PROPERTY CXX_STANDARD 17)
-target_link_libraries(gemma hwy hwy_contrib sentencepiece)
+target_link_libraries(gemma hwy hwy_contrib sentencepiece stdc++fs)
```
friendlyanon, line 67: This is a toolchain requirement only for some old compilers. Either update the compiler, or pass in the link flag from the command line/a toolchain file/a preset via [CMAKE_EXE_LINKER_FLAGS](https://cmake.org/cmake/help/latest/variable/CMAKE_EXE_LINKER_FLAGS.html).

friendlyanon (commented):
This shouldn't be part of the project code. Toolchain requirements like this must be satisfied on the user's end, not on the project side.

PR #34, opened 2024-02-24
reset conversation

kishida:
reset conversation when `%c` or `%C` is typed


austinvhuang:
Thanks, this is handy. I think we'll default to `--multiturn 0`, there are some bugs with continuations that are being worked out. Also eventually we'll want a proper parser to deal with special commands.

PR #36, opened 2024-02-24
Rename BUILD to BUILD.bazel. 

dan-zheng:
From https://bazel.build/concepts/build-files:

> Note: The `BUILD` file can be named either `BUILD` or `BUILD.bazel`. If both files exist, `BUILD.bazel` takes precedence over `BUILD`. For simplicity's sake, the documentation refers to these files simply as `BUILD` files.

This fixes an error on macOS due to `build` and `BUILD` having conflicting names.

Enable `macos-latest` in GitHub Actions CI, to catch these errors in the future.


dan-zheng:
Merging since this hardens CI, which will be good for preventing further build regressions.

Also removes `BUILD` section that should have been stripped by Copybara.

jan-wassenberg (commented):
LGTM

Issue #52, opened 2024-02-26
GRPC support - in scope?


justinsb:
I'd like to be able to run gemma.cpp on kubernetes.  A first step in my rough plan is to add a client/server mode, and I thought I would add GRPC support.  Is the project open to having a `contrib` directory where we can collaborate on this sort of thing?  In future, I'm imagining we could put things like kubernetes manifests in there also.

I have a simple server (though my C++ is not good!) and an example client in golang which I will send as a WIP PR to make the discussion more concrete.

austinvhuang:
Hi @justinsb thanks for taking the initiative. There's been interest in client/server capabilities and I think there's some obvious use cases + value in that. 

There's a few things being worked out that intertwine with such an implementation:

1. The server piece of this would probably be its own frontend entrypoint (basically in-place of run.cc), how should these alternative frontends be implemented? (I was working on some example demos, but they're on pause while we're traiging this first wave of post-launch PRs/issues) 
2. Should these implementations live in this repo (eg contrib/) or separate repos (like https://github.com/namtranase/gemma-cpp-python)?
3. There's some minor decoupling code cleanup to better support gemma.cpp-as-a-library / alternative frontend use cases (eg we probably want to decouple the cli argument handling more cleanly).

My suggestion is keep your initial implementation light (some interfaces may change as a result of #3). Can use that as a basis for thinking through design gaps + cleanup needed. A meta q is where to have more involved design discussions w/ community (i've also opened up the discussions tab up top but haven't made use of it yet, may look into a discord).

justinsb:
> There's been interest in client/server capabilities and I think there's some obvious use cases + value in that.

That's great news, and I agree!
 
> There's a few things being worked out that intertwine with such an implementation:
> 
> 1. The server piece of this would probably be its own frontend entrypoint (basically in-place of run.cc), how should these alternative frontends be implemented? (I was working on some example demos, but they're on pause while we're traiging this first wave of post-launch PRs/issues)

I had a go, my alternative entrypoint `server.cc` basically copies and pastes run.cc and has started swapping out functionality so that it read & writes over GRPC instead of stdin/stdout.  I will give it a cleanup pass, but I think the core gemma.cpp is very amenable to reuse already and there's not a ton of boilerplate between run.cc and server.cc, so that's a good sign that this is well architected IMO.

> 2. Should these implementations live in this repo (eg contrib/) or separate repos (like https://github.com/namtranase/gemma-cpp-python)?

[...]

PR #53, opened 2024-02-26
WIP: Add grpc server to contrib

justinsb:
As described in #52, a very rough GRPC server with a golang example client, with a goal of working towards running on kubernetes with gemma.cpp running in its own Pods.

This may or may not be in scope, but I want to make #52 more concrete!


austinvhuang:
Working on carving out a path towards a dedicated repository. Closing for now but will post here for follow-ups/discussion.

PR #58, opened 2024-02-27
clean up formatting after 129e66ada2b4e461bdf28b88b70cd2465cb213e4

austinvhuang:
- clean up formatting after 129e66ada2b4e461bdf28b88b70cd2465cb213e4
- add .clang-format defaults
- minor updates to DEVELOPERS doc


DEVELOPERS.md
```diff
@@ -71,6 +71,18 @@ The implementation code is roughly split into 4 layers, from high to low level:
 4.  Backend (`highway`) - Low-level hardware interface (SIMD in the case of
     highway) supporting the implementations in (3).
 
+Besides these layers, supporting utilities are:
+
+- `compression/` - model compression operations. the 8-bit switched floating
```
dan-zheng, line 76:
```suggestion
- `compression/` - model compression operations. The 8-bit switched floating 
```

dan-zheng (approved):
Looks great! Small nit only.

.clang-format
dan-zheng, line 1: As @pculliton suggests: it would be good to look into Github Actions enforcement of clang-format.

(https://github.com/marketplace/actions/clang-format-github-action)

dan-zheng, line 1: Is this .clang-format file based on `clang-format -style=Google`? If not, what are the differences?

Following `-style=Google` seems like a good idea to me, though I'm not up-to-date with the details.

austinvhuang: good point, replicated some settings but `-style=Google` is the right solution (there are some subtle diffs). Adding a commit that dumps `-style=Google` settings as is.

jan-wassenberg: hm, the Highway one merely sets BasedOnStyle:Google in the style file. That is quite close to the internal IDE. Why not just go with that, rather than output all settings via script and set them again explicitly?

DEVELOPERS.md
```diff
@@ -83,6 +83,14 @@ A `.clang-format` configuration is provided with our defaults, please run source
 files through `clang-format` (or a formatter that produces equivalent behavior)
 before finalizing PR for submission.
 
+The `.clang-format` is the google style (as of feb 27 2024), except with 
+`SortIncludes` set to `false` to avoid breaking copybara path substitutions
```
jan-wassenberg, line 87: Sorting is generally desirable. An alternative is to `// clang-format off` only where it would break us, what do you think?

gemma.h
```diff
@@ -29,10 +29,10 @@
 // copybara:import_next_line:gemma_cpp
 #include "configs.h"  // kSeqLen
 // copybara:import_next_line:gemma_cpp
+#include "util/args.h"  // ArgsBase
```
jan-wassenberg, line 32: I think an even simpler solution to the sorting issue is to add a comment line after the this line, which serves as a 'barrier' it does not sort around.

austinvhuang: there was actually a bottom barrierin @dan-zheng 's original copybara transformations. i think it might've been taken out to simplify since it was technically redundant, but if it also serves to mitigate linter breakage maybe it's worth reintroducing. for now i'll just add a comment line in manually for these cases.

DEVELOPERS.md
austinvhuang: i believe that would almost work but i think there would still be corner cases that can slip through, eg a PR introduces a new #include. I can't think of anything that's entirely leak proof atm. For now, I'll use your barrier suggestion re-enable sorting.

.clang-format
austinvhuang: The underlying issue is that the style linter source transformation is different on either side of the copybara source transformation. Disabling include sorting makes the style inter source transformation invariant to the copybara source transformation, with the goal of guaranteeing that style linter application can't break the build as it did in this case.

The specification is small -style="{BasedOnStyle: Google, SortIncludes: false}" and the shell script isn't strictly necessary to include, it's mostly just a reminder so the dumped configuration is reproducible.

I'm good with the comment barrier solution proposed though. We'd be modifying the source so the linter doesn't break the build instead of the other way around.

jan-wassenberg: I understand the issue with the include sorting. My comment is more along the lines of "should we include all of the style spec rather than just say BasedOn:Google". I agree with your point that a complete spec is more reproducible in the face of various clang-format versions which do exist externally, so the script does make sense :)

DEVELOPERS.md
jan-wassenberg: Right, I like the idea of a next_line/include/end atomic unit. One caveat there is that any // comment after include would break the build, which is a concern.

PR #68, opened 2024-02-29
Add OpenCL Wrapper Codes

Rliop913:
Opencl Wrapper for GPU implementation


austinvhuang:
Hi @Rliop913 thanks for taking the initiative, before getting too deep into this:

- we haven't decided what the approach is with accelerator support. currently leaning more towards webgpu than opencl given the trends of support for those APIs
- if there's an SDK that needs to be vendored, would probably prefer to do it via submodules than to have a copy in the repo.

if you have any thoughts/comments feel free to post here, might be useful to discuss synchronously as well (may setup a discord in the near future).

PR #74, opened 2024-03-01
Add instructions to download from Hugging Face Hub

osanseviero:
Hi there :wave: In collaboration with Google, we added the gemma.cpp files to the Hugging Face Hub. This PR adds instructions to manually or programmatically download the needed files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,11 +78,12 @@ Alternatively, visit the [gemma.cpp](https://huggingface.co/models?other=gemma.c
 models on the Hugging Face Hub. First go the the model repository of the model of interest
 (see recommendations below). Then, click the `Files and versions` tab and download the 
 model and tokenizer files. For programmatic downloading, if you have `huggingface_hub`
-installed, you can also run:
+installed, you can also download by running:
 
 ```
-huggingface-cli download google/gemma-2b-cpp --local-dir build/
-``` 
+huggingface-cli login # Just the first time
+huggingface-cli download google/gemma-2b-sfp-cpp --local-dir build/
+```
 
 2B instruction-tuned (`it`) and pre-trained (`pt`) models:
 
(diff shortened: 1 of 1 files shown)


filipeborato (approved):

README.md
```diff
@@ -65,15 +65,25 @@ winget install --id Kitware.CMake
...
 `-sfp` checkpoints.
 
+Alternatively, visit the [gemma.cpp](https://huggingface.co/models?other=gemma.cpp)
+models on the Hugging Face Hub. First go the the model repository of the model of interest
+(see recommendations below). Then, click the `Files and versions` tab and download the 
+model and tokenizer files. For programmatic downloading, if you have `huggingface_hub`
+installed, you can also run:
+
+```
+huggingface-cli download google/gemma-2b-cpp --local-dir build/
```
austinvhuang, line 84: Few minor tweaks:

- Use `google/gemma-2b-sfp-cpp` as the example here instead of `google/gemma-2b-cpp`
- Add a comment about `login`, I ran into the authentication error the first time I ran this command.
- Can you add a `models/` directory with a `.gitignore` placeholder and use that as the example local directory? Want to nudge people towards that going forwrd since `build/` gets wiped when changing configurations.

README.md
```diff
@@ -96,7 +106,7 @@ weights enable faster inference. In general, we recommend starting with the
 > **Important**: We strongly recommend starting off with the `2b-it-sfp` model to
 > get up and running.
 
-### Step 2: Extract Files
+### Step 2: Extract Files (if downloading from Kaggle)
 
 After filling out the consent form, the download should proceed to retrieve a
 tar archive file `archive.tar.gz`. Extract files from `archive.tar.gz` (this can
```
austinvhuang, line 112: Add a comment like "If you downloaded from huggingface, skip to step 3".

austinvhuang (changes requested):
Thanks!

I'm assuming there's a google point-of-contact for these artifacts if we need to update them? (can email offline if we need to discuss specifics)

README.md
osanseviero: Done the first two points. I'm downloading the model to `build/` as that's where the models are moved in Kaggle. Should we update all the instructions from `build` to `models` in a follow-up PR?

austinvhuang: That would be helpful thanks, yes kaggle and HF instructions should probably both point to a directory that's not `build/`.

osanseviero: Should we merge this PR ahead and then I open a follow-up PR? I can also open a PR against this PR, but I think the first would be cleaner

osanseviero:
Regarding Google point of contact, Tris from DM and Manvinder Singh from Cloud can help

austinvhuang (approved):

PR #76, opened 2024-03-04
decoupling from args.h

Rliop913:
I don't know if this is the right way to decoupling, but for now, args.h has almost been disconnected, and I've finished debugging to the point where it runs locally.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,7 +46,6 @@ set(SOURCES
 if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "Release")
 endif()
-
 # Allowable types for WEIGHT_TYPE:
 # float - slow, not recommended
 # hwy::bfloat16_t - bfloat16 as implemented by https://github.com/google/highway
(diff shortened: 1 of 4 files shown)


austinvhuang:
Hi @Rliop913 - thanks for taking a stab this.

When I mentioned decoupling the core library from args.h, the underlying motivation is a separation of concerns. One should be able to use the library passing in the initialized state needed for computation (for compressed weights, tokenizer, etc.) without regard to how that state was constructed. In some frontends, it might be specified as a CLI arg, in another case, maybe the application only supports a specific model and the model choice is hard coded, in another case, maybe these objects are cloned between models. The library provides tools to read in args, construct objects, and compute over objects. However, the goal is for the latter to not care about the first two concerns (at least up to some layer of the stack, possibly GemmaImpl and GenerateImpl, but I haven't finalized the specifics).
 
In this implementation, rather than decouple this, it seems like it partially replicates functionality in args.h but implemented a bit differently in gemma_consts and Gemma_args. This is fair - I hadn't elaborated on the objective above in detail - but doesn't address the goal of refactoring.

Let me make a bit more progress on this branch first to try to clarify things and if you want to collaborate, we can coordinate ore in the discord chat.

PR #77, opened 2024-03-05
Use highway simd for SquaredL2 calculation

enum-class:
use highway for SquaredL2 calculation.

### DIFF
--- a/ops.h
+++ b/ops.h
@@ -340,11 +340,21 @@ static HWY_NOINLINE HWY_MAYBE_UNUSED float Dot(const float* HWY_RESTRICT a,
 // = Dot(a, a, size), but that is not allowed due to HWY_RESTRICT.
 static HWY_NOINLINE HWY_MAYBE_UNUSED float SquaredL2(
     const float* HWY_RESTRICT a, size_t size) {
-  float total = 0.f;
-  for (size_t i = 0; i < size; ++i) {
-    total += a[i] * a[i];
+  const hn::ScalableTag<float> d;
+  const size_t N = hn::Lanes(d);
+  HWY_DASSERT(size >= N);
+  HWY_DASSERT(size % (2 * N) == 0);
+
+  auto sum0 = hn::Zero(d);
+  auto sum1 = hn::Zero(d);
+  for (size_t i = 0; i + 2 * N <= size; i += 2 * N) {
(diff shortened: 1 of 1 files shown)


ops.h
```diff
@@ -340,11 +340,21 @@ static HWY_NOINLINE HWY_MAYBE_UNUSED float Dot(const float* HWY_RESTRICT a,
 // = Dot(a, a, size), but that is not allowed due to HWY_RESTRICT.
 static HWY_NOINLINE HWY_MAYBE_UNUSED float SquaredL2(
     const float* HWY_RESTRICT a, size_t size) {
-  float total = 0.f;
-  for (size_t i = 0; i < size; ++i) {
-    total += a[i] * a[i];
+  const hn::ScalableTag<float> d;
+  const size_t N = hn::Lanes(d);
+  HWY_DASSERT(size >= N);
```
jan-wassenberg, line 345: Let's check `>= 2*N`, the loop step size.

ops.h
```diff
@@ -340,11 +340,21 @@ static HWY_NOINLINE HWY_MAYBE_UNUSED float Dot(const float* HWY_RESTRICT a,
...
+  HWY_DASSERT(size % (2 * N) == 0);
+
+  auto sum0 = hn::Zero(d);
+  auto sum1 = hn::Zero(d);
+  for (size_t i = 0; i + 2 * N <= size; i += 2 * N) {
```
jan-wassenberg, line 350: I used to like this loop structure but GCC raises static analysis warnings about it due to overflow if (theoretically) `i` gets huge.
The safer alternative is to check `if (size >= 2*N)` (or assert, as you are already doing), then `for (i = 0; i <= size - 2*N; i += 2*N)`. Does that make sense?

jan-wassenberg (changes requested):
Nice, thanks for vectorizing :)

jan-wassenberg (approved):
Thanks for making the change!

enum-class:
Same issue with copybara here, I think it needs a restart too:)

austinvhuang:
Since https://github.com/google/gemma.cpp/pull/78 was merged this may need to merge the updated `dev` into this branch to be able to finalize the merge. May take a stab tomorrow if nobody else gets to it.

austinvhuang:
Thanks, merged now!

PR #78, opened 2024-03-05
use hwy/simd for RMSNorm(f, bf, f) calculation

enum-class:
use highway/simd for RMSNorm(f, bf, f) calculation for performance purpose.

```
static HWY_NOINLINE HWY_MAYBE_UNUSED void RMSNorm(
     const float* HWY_RESTRICT x, const hwy::bfloat16_t* HWY_RESTRICT weight,
     float* HWY_RESTRICT out, size_t size) ;
```

### DIFF
--- a/ops.h
+++ b/ops.h
@@ -362,12 +362,30 @@ static HWY_NOINLINE HWY_MAYBE_UNUSED void RMSNorm(
 static HWY_NOINLINE HWY_MAYBE_UNUSED void RMSNorm(
     const float* HWY_RESTRICT x, const hwy::bfloat16_t* HWY_RESTRICT weight,
     float* HWY_RESTRICT out, size_t size) {
+  namespace hn = hwy::HWY_NAMESPACE;
+
   constexpr float eps = 1e-6f;
-  float ss = SquaredL2(x, size);
-  ss = 1.0f / sqrtf(ss / StaticCast<float>(size) + eps);
-  for (size_t j = 0; j < size; j++) {
-    // Note 1.0f centering here
-    out[j] = (1.0f + hwy::F32FromBF16(weight[j])) * (ss * x[j]);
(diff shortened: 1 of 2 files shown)


ops.h
```diff
@@ -362,12 +362,30 @@ static HWY_NOINLINE HWY_MAYBE_UNUSED void RMSNorm(
...
     float* HWY_RESTRICT out, size_t size) {
+  namespace hn = hwy::HWY_NAMESPACE;
+
   constexpr float eps = 1e-6f;
-  float ss = SquaredL2(x, size);
-  ss = 1.0f / sqrtf(ss / StaticCast<float>(size) + eps);
-  for (size_t j = 0; j < size; j++) {
-    // Note 1.0f centering here
-    out[j] = (1.0f + hwy::F32FromBF16(weight[j])) * (ss * x[j]);
+  constexpr size_t unroll_size = 2;
```
jan-wassenberg, line 368: For constexpr please use kUnrollSize naming :)

jan-wassenberg (changes requested):
Very nice, thanks for implementing! One style suggestion:

ops.h
enum-class: I 've set up Clang-Tidy to enforce the kCamelCase style for constexpr variables in the codebase, aiming for automated style checking and corrections.
Maybe in future it can be applied to fix all constexpr styles

jan-wassenberg (approved):
Nice, great that you have updated the lint to check for this :)

enum-class:
It seems copybara has faced an issue, maybe the rerun can solve it :))

jan-wassenberg:
Yes indeed, still have an issue with copybara transforms. Looking into it :)

jan-wassenberg:
Issue resolved 👍 One more internal review, then it will be merged.

PR #82, opened 2024-03-06
libgemma API refactor - decouple from interactive repl demo specifics, add hello world example using libgemma

austinvhuang:
This PR refactors the library interfaces to facilitate multiple frontends / apps to make better use of libgemma. It removes coupling between the specific interactive interface of `run.cc`, it's set of CLI arguments and the core library API in `gemma.h` and `gemma.cc`. As a test of this library use case, we add a minimal usage example in `examples/hello_world/` with minimal setup to interact with the library interfaces.

- Decouple Gemma, GemmaImpl, GemmaInterface, Generate, from CLI arguments
- Add a simple "hello world" library example
- Additional convenience functions / types to reduce verbosity
- This branch also includes PR https://github.com/google/gemma.cpp/pull/81 from @ufownl decoupling KV cache from GemmaImpl.

Minor:
- Add examples/ directory and README for examples/hello_world
- Fix LoaderArgs validaiton order
- Minor computation quality tweaks
- Fix up indentation, line breaks in --help message

See also:
- Discussion in the KV Cache PR: https://github.com/google/gemma.cpp/pull/81#discussion_r1514769338


examples/hello_world/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+
+
+# Allow for both local and remote building)
+option(BUILD_MODE "'local' or 'remote' git fetch for builds")
+if (NOT BUILD_MODE)
+  set(BUILD_MODE "remote")
+endif()
+if (BUILD_MODE STREQUAL "local")
+  FetchContent_Declare(gemma SOURCE_DIR ../../..)
```
jan-wassenberg, line 34: Not only two dirs up? Aren't we in examples/hello_world?

examples/hello_world/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+  FetchContent_Declare(gemma GIT_REPOSITORY https://github.com/google/gemma.cpp.git GIT_TAG 8c7b2cf61b9794b806de091685dc6739dd3db837)
+endif()
+FetchContent_MakeAvailable(gemma)
+
+if(NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE "Release")
+endif()
+
+add_executable(hello_world run.cc)
+set_property(TARGET hello_world PROPERTY CXX_STANDARD 17)
```
jan-wassenberg, line 45: Still required even if we set it for the project?

examples/hello_world/README.md
```diff
@@ -0,0 +1,41 @@
+# Hello World Example
+
+This is a minimal/template project for using `gemma.cpp` as a library. Instead of an interactive interface, it sets up the model state and generates text for a single hard coded prompt.
+
+Build steps are similar to the main `gemma` executable. From inside the top-level directory. For now only `cmake`/`make` is available for builds (PRs welcome for other build options).
```
jan-wassenberg, line 5: Can clarify what is meant by "from inside the top-level directory"?

examples/hello_world/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+As with the top-level `gemma.cpp` project you can use the `make` commands `-j` flag to use parallel threads for faster builds.
+
+From inside the `gemma.cpp/examples/hello_world/build` directory, there should be a `hello_world` executable. You can run it with the same 3 model arguments as gemma.cpp specifying the tokenizer, compressed weights file, and model type, for example:
+
+```sh
+./hello_world --tokenizer tokenizer.spm --compressed_weights 2b-it-sfp.sbs --model 2b-it
+```
+
+Should  print a greeting to the terminal:
```
jan-wassenberg, line 37: extra space after Should

examples/hello_world/run.cc
```diff
@@ -0,0 +1,83 @@
...
+// copybara:import_next_line:gemma_cpp
+#include "gemma.h"
+// copybara:end
+// copybara:import_next_line:gemma_cpp
+#include "util/args.h"
+// copybara:end
+#include "hwy/contrib/thread_pool/thread_pool.h"
+
+std::vector<int> tokenize(
+    std::string prompt_string,
```
jan-wassenberg, line 27: Extra copy of the arg. How about const&, then created a new std::string wrapped = "sot" + prompt + "eot"?

gemma.h
```diff
@@ -192,19 +182,50 @@ struct InferenceArgs : public ArgsBase<InferenceArgs> {
...
+                   size_t start_pos, KVCache& kv_cache, hwy::ThreadPool& pool,
+                   hwy::ThreadPool& inner_pool, const StreamFunc& stream_token,
+                   const AcceptFunc& accept_token, std::mt19937& gen,
                    int verbosity);
 
+// Convenience function for the common case:
+// - Bundle runtime parameters as RuntimeConfig
+// - No threadpools within threadpools (inner_pool = dummy)
+// - All tokens accepted
+void GenerateGemma(Gemma& gemma, RuntimeConfig runtime_config,
```
jan-wassenberg, line 224: A simplified interface seems useful. Maybe we should also move stream_token and gen inside RuntimeConfig? (Which we should pass by ref = RuntimeConfig&)

run.cc
```diff
@@ -191,7 +190,8 @@ void ReplGemma(gcpp::Gemma& model, hwy::ThreadPool& pool,
       }
     }
 
-    HWY_ASSERT(model.Tokenizer().Encode(prompt_string, &prompt).ok());
+    // HWY_ASSERT(model.Tokenizer().Encode(prompt_string, &prompt).ok());
```
jan-wassenberg, line 193: Delete commented out line?

gemma.cc
```diff
@@ -231,43 +231,62 @@ struct Activations {
...
+      return CreateKVCache<ConfigGemma7B>();
+    default:
+      HWY_ABORT("Model type %d unknown.", static_cast<int>(type));
+  }
+}
+
 template <class Config>
 struct GemmaImpl : public GemmaInterface {
-  GemmaImpl(const LoaderArgs& args, hwy::ThreadPool& pool);
+  GemmaImpl(  // const LoaderArgs& args,
```
jan-wassenberg, line 264: Remove commented out line?

gemma.cc
```diff
@@ -729,75 +763,101 @@ KVCache CreateKVCache(size_t size_cache_pos, size_t seq_len) {
...
 
 template <class Config>
-GemmaImpl<Config>::GemmaImpl(const LoaderArgs& args, hwy::ThreadPool& pool)
-    : compressed_weights(
-          HWY_DYNAMIC_DISPATCH(GetCompressedWeightsT)(args, pool)),
+GemmaImpl<Config>::GemmaImpl(
+    std::unique_ptr<sentencepiece::SentencePieceProcessor>& tokenizer,
+    hwy::AlignedFreeUniquePtr<uint8_t[]>& compressed_weights,
+    hwy::ThreadPool& pool)
+    // GemmaImpl<Config>::GemmaImpl(const LoaderArgs& args, hwy::ThreadPool&
```
jan-wassenberg, line 770: Remove commented out line?

gemma.cc
```diff
@@ -729,75 +763,101 @@ KVCache CreateKVCache(size_t size_cache_pos, size_t seq_len) {
...
+   kv_cache, pool, inner_pool, stream_token, accept_token, gen, verbosity);
 }
 
-Gemma::Gemma(const LoaderArgs& args, hwy::ThreadPool& pool) {
-  const Model model_type = args.ModelType();
-  model_training = args.ModelTraining();
+Gemma::Gemma(const Path& tokenizer_path, const Path& compressed_weights_path,
+             const Path& weights_path, Model model_type,
+             hwy::ThreadPool& pool) {
+  PROFILER_ZONE("Startup.tokenizer");
```
jan-wassenberg, line 807: This ZONE covers the entire basic block, until the next }. We should structure the code like so:

```
tokenizer = make_unique;
{
  PROFILER_ZONE("Startup.tokenizer");
  HWY_ASSERT(tokenizer->Load(tokenizer_path.path).ok());
}
```
Note the braces around the code.

jan-wassenberg (changes requested):
👍

examples/hello_world/CMakeLists.txt
jan-wassenberg: Got it. Can consider adding a comment because it was surprising.

gemma.h
austinvhuang: I was thinking of this too, but decided for now to make stateful things more explicit for now. For example, say a user who's new to the code wants to make the randomness reproducible - it's implied what to do from this interface - seed the RNG that gets passed in.

Packing it inside RuntimeConfig adds a level of indirection + opacity - first they have to know or guess that there's an rng that is getting passed in (and not say, initialized somewhere internally), then know to check the internals of RuntimeConfig to find it, then find the place in application codebase where RuntimeConfig is constructed, and finally make the change. 

Granted for us w'ere able to track this without too much cognitive overhead, but I want to try to keep things flatly legible as much as we can until other tradeoffs make it obviously needed+beneficial.

jan-wassenberg: Seems reasonable and fine for now :)

jan-wassenberg (approved):
👍

examples/hello_world/run.cc
```diff
@@ -0,0 +1,83 @@
...
+                          int token, float) {
+    ++pos;
+    if (pos < ntokens) {
+      // print feedback
+    } else if (token != gcpp::EOS_ID) {
+      std::string token_text;
+      HWY_ASSERT(tokenizer->Decode(std::vector<int>{token}, &token_text).ok());
+      std::cout << token_text << std::flush;
+    }
+    return true;
```
zeerd, line 72: I am trying to make a QT version of gemma, and find an issue about this return value.
The calling of stream_token split to two parts. first for `pos < ntokens`. second for token_text.
We can break the second part by return false here, but could not break the first part when given a large input (ex. a file).
Hope someone can optimize it.

austinvhuang: What's QT in this context (quantized?) ?

I think the issue you're pointing out makes sense - will have a closer look at your PR.

zeerd: I am not sure if I understood your question...
I have a stop button to [break](https://github.com/zeerd/gemma.qt/blob/main/mainwindow.cpp#L201) the progress.

austinvhuang: I was asking about what QT stood for in this context, but I see you meant the Qt UI framework - this is very cool! I've used Qt in the past myself. 

In the future I may add some demos to try some unorthodox UX patterns for interacting with an LLM, but they will probably be in the form of ftxui prototypes. They might be useful for you.

Look forward to see where your project goes.

PR #83, opened 2024-03-07
Auto-labels 'Gemma' on 'gemma' issues/PRs.

shmishra99:
This workflow automatically identifies issues and pull requests (PRs) related to `Gemma`.
It searches for the keyword `Gemma` (case-insensitive) in both the title and description of the issue/PR.
If a match is found, the workflow adds the label `Gemma` to the issue/PR.

On first run, the workflow will try to create the label `Gemma`. If unsuccessful, the repository owner will need to create it manually.

label: `Gemma`

### DIFF
--- a/.github/workflows/labeler.yaml
+++ b/.github/workflows/labeler.yaml
@@ -1,0 +1,42 @@
+# Copyright 2024 Google LLC. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


austinvhuang:
Hi @shmishra99 the repo is gemma.cpp so on one hand every issue could be tagged as gemma related. On the other hand filtering for "gemma" as a keyword is not going to get additional specificity of whether the issue is about gemma the model or some other part of the repo.

There's already some amount of workflow automation complexity to this repo between copybara + CI automation, so we're trying not to add more than is needed. If I'm missing context here, feel free to ping me offline directly.

PR #94, opened 2024-03-12
Make stream_token can break prefill, too

zeerd:
When we give the gemma a large input(ex. a file), we need a way to stop it.

### DIFF
--- a/gemma.cc
+++ b/gemma.cc
@@ -25,12 +25,12 @@
 #include "compression/compress-inl.h"
 // copybara:import_next_line:gemma_cpp
 #include "ops.h"
-// copybara:import_next_line:gemma_cpp
-#include "util/args.h"  // Path
 #include "hwy/contrib/matvec/matvec-inl.h"
 #include "hwy/highway.h"
 #include "hwy/profiler.h"
 #include "hwy/timer.h"
+// copybara:import_next_line:gemma_cpp
+#include "util/args.h"  // Path
 // copybara:import_next_line:sentencepiece
 #include "src/sentencepiece_processor.h"
 // copybara:end
@@ -43,11 +43,12 @@
 #include <math.h>  // sqrtf
 #include <stddef.h>
 #include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
 
 #include <algorithm>
 #include <array>
 #include <cmath>
-#include <cstdlib>
 #include <filesystem>  // NOLINT
 #include <iostream>
 #include <memory>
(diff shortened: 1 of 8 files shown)


gemma.cc
```diff
@@ -529,7 +529,7 @@ void GenerateImpl(GemmaImpl<TConfig>& gemma, const InferenceArgs& args,
 
   // Prefill stops before prompt.size() - 1 since the last prompt token is the
   // first input token for generation.
-  while (pos_offset < prompt.size() - 1) {
+  while (pos_offset < prompt.size() - 1 && token != EOS_ID) {
```
austinvhuang, line 532: Is there a reason to override the token value vs. track the stream_token return state with its own bool variable? Keeping the two forms of state seems easier to reason about rather than having to track "Is token EOS_ID because that's what was generated or because stream_token was false..."

austinvhuang (changes requested):
Thanks, the need to be able to exit makes sense. 

Probably beyond the scope of this PR, but I wonder if in the long term this starts to get messy trying to track this in stream_token It's certainly doable, but there's probably a breaking point where stream_token is trying to accomplish too much control flow logic.

There's some related design choices around customizable hooks/callbacks in the context of trainer abstractions (eg in pytorch lightning), might look to see how other frameworks deal with this.

Specific tactical items to get this merged:

- Can you use a separate variable to track the stream_token return value instead of overwriting token? (unless there's a rationale I'm not seeing)
- Also, may need to merge/test with the latest state of the `dev` branch.

gemma.cc
zeerd: You are right. I want to change as little code as possible. But it's not meaningful.
I will think more again.

zeerd:
In fact, I don't like returning in the middle of a function.
However, it is strange to enclose the second half of the entire function with conditional judgments.

Another side, I recommend splitting GenerateImpl into two functions.
But this might be a big changing.

jan-wassenberg:
Yes, GenerateImpl does not fit on a single screen and I agree it would make sense to split into two functions.
We just did that with Attention as well. Would you prefer lambdas or actual functions? A function probably only makes sense if there are relatively few parameters.

zeerd:
I have less knowlage about neurol network.
But I am thinking about an usecase :

There is a source file, which has 5 functions.I want the gemma help me to know what the 3rd function dose.

We know that, if the gemma read the whole file first, it could give the summarization of the 3rd function better.

So, maybe I could do:

* 1st, let gemma read the whole file to summarize but not output(to save time time, as I do not care it).
* 2nd, let gemma read the 3rd function again , summarize and output.

I want to know, If we splite the GenerateImpl to two part. Could them help me to implement this case?

If the answer is not (or this could not help me to save time), I think lambda is enough.

PR #95, opened 2024-03-12
Convert PyTorch checkpoints to gemma.cpp checkpoints

pculliton:
Adds script for conversion of PyTorch to gemma.cpp formatted weights.

Tested: converted 2B and 7B PyTorch checkpoints and evaluated their output.

Key features:
- Layer transpositions per specs, based on [gemma_pytorch](https://github.com/google/gemma_pytorch).
- Handling of MQA->MHA conversion in 2B models.
- Transposition checking with failure reporting prior to writing.
- Customizable args.

### DIFF
--- a/util/convert_weights.py
+++ b/util/convert_weights.py
@@ -1,0 +1,55 @@
+# WIP - DO NOT MERGE
+
+import torch
+from gemma import config
+from gemma import model as gemma_model
+import numpy as np
+
+
+def param_names():
+    """Return parameter names in the order they are expected for deserialization."""
+    names = ["embedder.weight", "model.norm.weight"]
+    # note *weight_scaler params are ignored in the forward computation unless quantization is being used.
+    # since we are working with the full precision weights as input, don't include these in the parameters being iterated over
+    layer_params = [
+        "self_attn.qkv_proj.weight",  # attn_vec_einsum_w
+        "self_attn.o_proj.weight",  # qkv_einsum_w
+        "mlp.gate_proj.weight",  # qkv_einsum_w
+        "mlp.up_proj.weight",  # gating_einsum_w
+        "mlp.down_proj.weight",  # linear_w
+        "input_layernorm.weight",  # pre_attention_norm_scale
+        "post_attention_layernorm.weight",  # pre_ffw_norm_scale
+    ]
+    for layer in range(18):
+        for layer_param in layer_params:
+            names = names + ["model.layers." + str(layer) + "." + layer_param]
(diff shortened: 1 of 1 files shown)


util/convert_weights.py
```diff
@@ -0,0 +1,216 @@
...
+
+from collections import defaultdict
+import torch
+from gemma import config
+from gemma import model as gemma_model
+import numpy as np
+import argparse
+import os
+
+# WIP - DO NOT MERGE
```
jan-wassenberg, line 25: Still WIP?

jan-wassenberg (approved):
Nice! Do we want to merge to main, or dev?

austinvhuang:
> Nice! Do we want to merge to main, or dev?

The automation will land it in dev automatically. We can push it to main right away if we want, usually i'll do a quick test to see that text generation says something reasonable but in this case the code is isolated.

austinvhuang:
Just checking, does this write a float32 .sbs that would be read in via `--weights` and then compressed (even to bf16)? 

If so, we should provide some detailed notes on usage - we mostly skip the `--weights` -> `--compressed_weights` workflow in the docs so far since the provided checkpoints are already compressed. We'll need to add notes for people to go through that flow for the first time.

Doesn't have to be in this PR, but we should add a README in this doc (or alternatively print the instructional text as part of tis script).

jan-wassenberg:
f32 .sbs can only be loaded via --compressed_weights. Once this lands, I can add a compressed_weights -> out_weights path that changes the compression. Then we can entirely remove --weights.

austinvhuang (approved):
Very nice. I like how simple this turned out compared to ... our earlier incarnations :)

We don't have to do it for this PR, but we should provide instructions on usage in the near future. Esp the expectations around model compression workflow I don't think most people will infer without a guide.

Two possibilities to make the compression workflow more polished - 1) factor out a small C++ tool that's *just* responsible for compression instead of crowding the CLI arguments of the `gemma` CLI. Alternatively 2) implementing subcommand hierarchy if we want to go the monolithic command route would be cleaner.

These both take some effort and are probably less high priority than other low hanging fruit, but they are medium term QoL tweaks to consider.

dan-zheng:
Closing manually since copybara-service seems to have merged commits from this PR into `dev` branch.

(Not sure why copybara-service didn't close the PR automatically)

util/convert_weights.py
abrahem79: > Still WIP?

PR #98, opened 2024-03-14
Add missing log that point to a failed Generation

zeerd:
if `pos >= max_tokens`, the next `for` will not be run.


jan-wassenberg (approved):
Thanks!

PR #105, opened 2024-03-19
Enhancing Utility Functions in ops.h

enum-class:
I have tried to use highway library in the AddFrom, MulBy, MulByConst, MulByConstAndAdd, and create_distribution functions.


ops.h
```diff
@@ -551,47 +551,66 @@ static HWY_NOINLINE HWY_MAYBE_UNUSED void RopeAndMulBy(const float mul,
...
+static HWY_NOINLINE void MulByConst(const float c, float* HWY_RESTRICT x,
+                                    const size_t size, const size_t max_pos) {
   HWY_DASSERT(max_pos <= size);
-  for (size_t i = 0; i < max_pos; ++i) {
-    x[i] *= c;
-  }
+  namespace hn = hwy::HWY_NAMESPACE;
+  using D = hn::ScalableTag<float>;
+  const D d;
+  const auto constant = hn::Set(d, c);
```
jan-wassenberg, line 589: Unfortunately this breaks on SVE, current compilers do not allow capturing references to vectors. How about passing in the `float c` and calling Set inside the lambda? Would be good to verify via gcc.godbolt.org that the codegen is still reasonable i.e., the compiler is hoisting Set out of the loop.

If not, we can consider adding TransformWithConstant or a similar function that passes the scalar constant in without a lambda capture, and passes the result of Set to the lambda. If we go that route, to ensure the code still works with older Highway versions, we could add a Highway version check and `#if` too old, then fall back to the previous plain C++ code.

ops.h
```diff
@@ -551,47 +551,66 @@ static HWY_NOINLINE HWY_MAYBE_UNUSED void RopeAndMulBy(const float mul,
...
+                                          float* HWY_RESTRICT out,
+                                          const size_t size,
+                                          const size_t max_pos) {
+  namespace hn = hwy::HWY_NAMESPACE;
+  using D = hn::ScalableTag<float>;
+  const D d;
+  const auto constant = hn::Set(d, c);
+  hn::Transform1(
+      d, out, max_pos, x,
+      [&constant](const auto d, const auto out_element, const auto x_element)
```
jan-wassenberg, line 612: Same issue here.

ops.h
```diff
@@ -693,15 +712,18 @@ template <size_t k>
...
-  denominator = 1.0f / denominator;
-  MulByConst(denominator, top_k.data(), k);
+  namespace hn = hwy::HWY_NAMESPACE;
+  using D = hn::ScalableTag<float>;
+  const D d;
+
+  const auto one = hn::Set(d, 1.0f);
+  const auto temperature_inv = hn::Div(one, hn::Set(d, temperature));
+
+  hn::Transform(d, top_k.data(), top_k.size(),
```
jan-wassenberg, line 722: And also here :) Either way, we can just compute 1.0f / temperature in scalar code and pass that to the lambda or the proposed TransformWithConstant (hopefully there is a shorter name).

jan-wassenberg (changes requested):
Nice, great to see the Transform* make this quite compact. Thanks for adding!

ops.h
enum-class: Thank you @jan-wassenberg , I had seen your comment [here](https://github.com/google/gemma.cpp/blob/8fb44ed6dd123f63dca95c20c561e8ca1de511d7/ops.h#L613)
But I have tried this code in [godbolt](https://gcc.godbolt.org/) with these flags `-O2 -march=armv8-a+sve` with Arm64 gcc-10.2 it was ok,
```

HWY_ATTR void MulByConst(const float c, float* HWY_RESTRICT x,
                                    const size_t size, const size_t max_pos) {
  HWY_DASSERT(max_pos <= size);
  namespace hn = hwy::HWY_NAMESPACE;
  using D = hn::ScalableTag<float>;
  const D d;
  const auto constant = hn::Set(d, c);
  hn::Transform(d, x, max_pos,
                [&constant](const auto d, const auto x)
                    HWY_ATTR { return hn::Mul(x, constant); });
}
``` 
Maybe I miss something.

Anyway, I will fix them :)

jan-wassenberg: Thanks for checking back. I see you are right, this indeed works on Clang and GCC. The key difference is the capture by ref - by value indeed fails to compile. We can keep your code as-is then :)

Would you also like to remove the comment you mentioned, and use copy-by-ref + Foreach there?

enum-class: Thanks , Yes sure, I will do it

jan-wassenberg (approved):
Super, the other spots are much nicer now, thank you :)

austinvhuang:
Nice! Did you see any performance differences (positive or negative) @enum-class ?

ops.h
```diff
@@ -600,79 +619,60 @@ static HWY_INLINE HWY_MAYBE_UNUSED void MulByConstAndAdd(
...
-      vmax = hn::Max(vmax, LoadU(d, x + idx));
-    }
-  }
-  vmax = hn::Max(vmax, LoadNOr(vmin, d, x + idx, mask_pos - idx));
-  vmax = hn::MaxOfLanes(d, vmax);  // broadcast
+  const auto vmin = hn::Set(d, hwy::LowestValue<float>());
+  auto vmax = vmin;
+  Foreach(d, x, mask_pos, vmin,
+          [&vmax](const auto d, const auto value)
+              HWY_ATTR { vmax = hn::Max(vmax, value); });
```
jan-wassenberg, line 635: @pchx noticed we are missing the reduction set here - after Foreach, broadcasting the max to all lanes:
`vmax = hn::MaxOfLanes(d, vmax); // broadcast`

ops.h
```diff
@@ -693,15 +693,18 @@ template <size_t k>
...
+  namespace hn = hwy::HWY_NAMESPACE;
+  using D = hn::ScalableTag<float>;
+  const D d;
+
+  const auto temperature_inv =
+      hn::Div(hn::Set(d, 1.0f), hn::Set(d, temperature));
+
+  hn::Transform(d, top_k.data(), top_k.size(),
+                [&temperature_inv](D d, hn::Vec<D> v) HWY_ATTR {
+                  return hn::Mul(hn::Exp(d, hn::Log(d, v)), temperature_inv);
```
jan-wassenberg, line 705: Looks like we have just v/temp here - I think we want log/temp, then only exp.
We are also missing the second half of the original code?

jan-wassenberg (changes requested):

ops.h
enum-class: Thanks @pchx and @jan-wassenberg.
In my local test I have noticed ReduceMax is missing.
I will fix it.

ops.h
enum-class: I am confused about vtemp things, could you please explain a bit more?

About the second half, I think it already exists inside the std::discrete_distribution.
I mean it will divide each weight by the sum of all n weights. am I missing something ?

enum-class:
> Nice! Did you see any performance differences (positive or negative) @enum-class ?

@austinvhuang Du to hardware limitations, I couldn't do the performance test on the entire pipeline.
I did check individual functions like MulBy, MulByConst and MulByConstAndAdd as much as fair I could.
Anyway, On 12th Gen Intel Core i7-1265U with SSE, -O3, I saw roughly 2x speedup.
I can share the details if you are interested:)

ops.h
jan-wassenberg: Sure! What I see now is exp(log(v)), which is just v, then we mul by 1/temp.
I think we intended exp(log*inv).

For the second half, I am not familiar with discrete_distribution. As a sanity check, how about we temporarily run both the old and the new code, and ensure their results are the same (which is generally a helpful methodology for porting to SIMD :) ), then we can remove the old before merging?

pchx: Ah, `discrete_distribution` does make the second half of the original code redundant -- we don't need to divide by the sum of weights.

Thanks!

jan-wassenberg (approved):
Impressive work! Thanks for adding the test :)

Issue #103, opened 2024-03-16
Weights converted from PyTorch don't seem to work properly


ufownl:
I followed the tutorial ["Fine-tuning Gemma Models in Keras using LoRA"](https://ai.google.dev/gemma/docs/lora_tuning) to fine-tune gemma-2b and converted the fine-tuned weights to PyTorch `ckpt`, which worked fine with gemma_pytorch and the output was as expected. Then I use the util script `convert_weight.py` to convert PyTorch `ckpt` to `sbs`, gemma.cpp can load `sbs` normally and generate compressed weights. However, the performance of the model output is far from gemma_pytorch, and occasionally it will output tokens that are not in the vocabulary, causing the tokenizer to return the error status. So I tried to look into it and found that the model structure configuration is different between gemma.cpp and gemma_pytorch.

This is the 2B size model structure of gemma.cpp:

```C++
struct ConfigGemma2B {
  static constexpr int kSeqLen = gcpp::kSeqLen;
  static constexpr int kVocabSize = 256128;
  static constexpr int kLayers = 18;
  static constexpr int kModelDim = 2048;
  static constexpr int kFFHiddenDim = 16 * 2048 / 2;  // = 16384
  static constexpr int kHeads = 8;
  static constexpr int kKVHeads = 8;   // TODO(austinvhuang): add MQA support
  static constexpr int kQKVDim = 256;  // query size == key size == value size
  static constexpr int kTopK = gcpp::kTopK;
};
```

This is the 2B size model structure of gemmpa_pytorch:

```Python
@dataclasses.dataclass
class GemmaConfig:
    # The number of tokens in the vocabulary.
    vocab_size: int = 256000
    # The maximum sequence length that this model might ever be used with.
    max_position_embeddings: int = 8192
    # The number of blocks in the model.
    num_hidden_layers: int = 28
    # The number of attention heads used in the attention layers of the model.
    num_attention_heads: int = 16
    # The number of key-value heads for implementing attention.
    num_key_value_heads: int = 16
    # The hidden size of the model.
    hidden_size: int = 3072
    # The dimension of the MLP representations.
    intermediate_size: int = 24576
    # The number of head dimensions.
    head_dim: int = 256
    # The epsilon used by the rms normalization layers.
    rms_norm_eps: float = 1e-6
    # The dtype of the weights.
    dtype: str = 'bfloat16'
    # Whether a quantized version of the model is used.
    quant: bool = False
    # The path to the model tokenizer.
[...]

PR #114, opened 2024-03-20
MQA Implementation for 2B models

ufownl:
This PR implements "Multi-Query Attention" for the 2B models and modifies vocabulary size to be the same as gemma_pytorch (mentioned in #103). It works fine with weights converted from gemma_pytorch **but will lead to the original gemma.cpp weights are unusable.**

It needs more testing, and I'll use it to test the fine-tuned weights.

### DIFF
--- a/gemma.cc
+++ b/gemma.cc
@@ -320,6 +320,16 @@ HWY_NOINLINE void Attention(size_t batch_start, size_t batch_idx, size_t layer,
 
   const size_t batch_offset = batch_idx * kModelDim;
 
+  auto ProjKV = [&](size_t k_offset, size_t v_offset, size_t kv_offset) {
+    TwoOfsMatVecLoop<kQKVDim, kModelDim>(
+        c_layer->c_qkv_einsum_w, k_offset, v_offset,
+        activations.pre_att_rms_out.data() + batch_offset,
+        kv_cache.key_cache.get() + kv_offset,
+        kv_cache.value_cache.get() + kv_offset);
+
+    Rope(kv_cache.key_cache.get() + kv_offset, kQKVDim, pos);
+  };
+
   pool.Run(0, kHeads, [&](const uint64_t head, size_t /*thread*/) HWY_ATTR {
     // linear projections to QKV
     constexpr const size_t head_offset =
@@ -339,13 +349,7 @@ HWY_NOINLINE void Attention(size_t batch_start, size_t batch_idx, size_t layer,
       const size_t kv_offset =
           pos * kCachePosSize + layer * kCacheLayerSize + head * kQKVDim;
 
-      TwoOfsMatVecLoop<kQKVDim, kModelDim>(
-          c_layer->c_qkv_einsum_w, k_offset, v_offset,
-          activations.pre_att_rms_out.data() + batch_offset,
-          kv_cache.key_cache.get() + kv_offset,
-          kv_cache.value_cache.get() + kv_offset);
-
-      Rope(kv_cache.key_cache.get() + kv_offset, kQKVDim, pos);
+      ProjKV(k_offset, v_offset, kv_offset);
     }
   });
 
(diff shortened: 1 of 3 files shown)


gemma.cc
```diff
@@ -311,47 +314,70 @@ HWY_NOINLINE void Attention(size_t batch_start, size_t batch_idx, size_t layer,
...
                                    batch_idx * kHeads * kQKVDim;
 
     Rope(q, kQKVDim, pos);
-    Rope(kv_cache.key_cache.get() + kv_offset, kQKVDim, pos);
     MulByConst(kQueryScale, q, kQKVDim);
     // Compute Q dot K scores
     for (size_t pos2 = 0; pos2 <= pos; ++pos2) {
       const size_t cache_offset =
-          pos2 * kCachePosSize + layer * kCacheLayerSize + head * kQKVDim;
+          kHeads == kKVHeads
```
austinvhuang, line 379: This is probably fine as long as we can verify the performance. Might consider turning head * kQKVDim into a const (eg head_offset) outside the for loop, setting it to 0 if kHeads == kKVHeads. That would get rid of conditional branching in the pos2 loop. tagging @jan-wassenberg in case you have other thoughts.

In the future as support is added for other model variations, we'll want to avoid branching in the inference path and have more flexibility for swapping in different inference codepaths for (but for MQA that's probably overkill)

gemma.h
```diff
@@ -75,9 +72,8 @@ struct GemmaInterface;
 
 struct Gemma {
   Gemma(const Path& tokenizer_path, const Path& compressed_weights_path,
-        const Path& weights_path, Model model_type, hwy::ThreadPool& pool);
-  Gemma(const Path& tokenizer_path, const Path& compressed_weights_path,
-        Model model_type, hwy::ThreadPool& pool);
+        const Path& weights_path, Model model_type, ModelTraining training,
```
austinvhuang, line 75: Don't have to tackle it in this PR, but we should consider consolidating Model and ModelTraining.

The issue is that as we add more variations, it doesn't always make sense to assume there's the cartesian product of all combinations, in which case having the ModelTraining metadata as part of the Model type might make more sense.

austinvhuang (commented):
Thanks very much, MQA is one of the most important low-hanging fruit to implement right now! Looks pretty good overall, have a look at the comment about avoiding branching.

Tagging @pculliton to check the model exporting + vocab size change and @jan-wassenberg on any perf suggestions.

gemma.h
austinvhuang: Now that the `experimental` changes were merged into `dev` (and `main`), you can go back to targeting PRs directly to `dev` (which will jan's commits).

ufownl: OK, I'll do this after moving the conditional branch out of the pos2 loop later.

gemma.cc
```diff
@@ -313,47 +314,72 @@ HWY_NOINLINE void Attention(size_t batch_start, size_t batch_idx, size_t layer,
...
 
-    TwoOfsMatVecLoop<kQKVDim, kModelDim>(
-        c_layer->c_qkv_einsum_w, k_offset, v_offset,
-        activations.pre_att_rms_out.data() + batch_offset,
-        kv_cache.key_cache.get() + kv_offset,
-        kv_cache.value_cache.get() + kv_offset);
+    ProjKV(k_offset, v_offset, kv_offset);
+  }
 
+  pool.Run(0, kHeads, [&](const uint64_t head, size_t /*thread*/) HWY_ATTR {
```
jan-wassenberg, line 366: if kHeads == kKVHeads it would be nice to stay in the same pool.Run rather than open up another, which does have a cost of several microseconds. What do you think of moving the contents of both pool.Run into functions, and then having either a big pool.Run or two halves (depending on MQA vs MHA)?

jan-wassenberg (changes requested):
Nice, thank you :) Some small suggestions:

gemma.cc
```diff
@@ -313,47 +314,72 @@ HWY_NOINLINE void Attention(size_t batch_start, size_t batch_idx, size_t layer,
...
 
-    const size_t batch_offset = batch_idx * kModelDim;
-
     MatVecLoop<kQKVDim, kModelDim>(
         c_layer->c_qkv_einsum_w, q_offset,
         activations.pre_att_rms_out.data() + batch_offset, q);
 
-    const size_t kv_offset =
-        pos * kCachePosSize + layer * kCacheLayerSize + head * kQKVDim;
+    if constexpr (kHeads == kKVHeads) {
```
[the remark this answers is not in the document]
ufownl: I initially added comments, but later I changed the structure and forgot to add them back. 😂

gemma.cc
ufownl: I agree with you, this will reduce the overhead of MHA, and the code readability should also be better.

austinvhuang (commented):
This LGTM, if the performance looks good/better (I'm curious how much) and generation looks correct + @jan-wassenberg LGTMs can probably move forward with merging to dev.

ufownl:
I tested the weights converted from gemma_pytorch (2b-it and 7b-it) and the generation looks fine.

jan-wassenberg (approved):
Very nice use of lambdas! Thanks for making the change.

PR #115, opened 2024-03-20
Added a missing space in app.h

villesundell:
When the user runs `--help`, they see the non-existent word `compressingnew`. This is because of a missing space, which is now added, resulting in `compressing new`.


jan-wassenberg (approved):
Thank you!

jan-wassenberg:
We have an issue with the copybara import mechanism. Working on it.

PR #116, opened 2024-03-21
fix: change hello_world cmake compiler use -std=c++17 

weedge:
use libgemma  util/app.h  this function `gcpp::LoaderArgs loader(argc, argv)` , Want to pass const char** to argv

### DIFF
--- a/examples/hello_world/run.cc
+++ b/examples/hello_world/run.cc
@@ -37,7 +37,7 @@ std::vector<int> tokenize(
   return tokens;
 }
 
-int main(int argc, char** argv) {
+int main(int argc, const char** argv) {
   gcpp::LoaderArgs loader(argc, argv);
 
   // Rough heuristic for the number of threads to use
--- a/examples/hello_world/run.cc
+++ b/examples/hello_world/run.cc
@@ -22,7 +22,7 @@
 #include "util/args.h"
 // copybara:end
 // copybara:import_next_line:gemma_cpp
-#include "util/app.h" // LoaderArgs
+#include "util/app.h"  // LoaderArgs
 // copybara:end
 #include "hwy/contrib/thread_pool/thread_pool.h"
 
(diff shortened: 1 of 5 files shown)


examples/hello_world/run.cc
```diff
@@ -37,7 +37,7 @@ std::vector<int> tokenize(
   return tokens;
 }
 
-int main(int argc, char** argv) {
+int main(int argc, const char** argv) {
```
jan-wassenberg, line 40: Why do we use const here? char** or char*[] is standard, see 3.6.1 in https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3337.pdf.

weedge: argv from command  params just read, so want use const char**,   want use libgemma util/app.h this function `gcpp::LoaderArgs loader(argc, argv)`  i want to cast const char** to  char** ,    like this `char** nonConstPtr = const_cast<char**>(constPtr);`;  but this command argv is immutable

weedge: > Why do we use const here? char** or char*[] is standard, see 3.6.1 in https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3337.pdf.

use `main(int argc, char** argv)` it's ok~

examples/hello_world/run.cc
```diff
@@ -38,7 +38,7 @@ std::vector<int> tokenize(
 }
 
 int main(int argc, char** argv) {
-  gcpp::LoaderArgs loader(argc, argv);
+  gcpp::LoaderArgs loader(argc, (const char**)argv);
```
jan-wassenberg, line 41: I still do not understand how this change is helpful? :) Const is generally helpful, but for historical reasons we are stuck with non-const for args. It seems like casting is worse than the safety benefit.
We can keep the c++17 change, but let's revert the const or at least move to a separate PR to discuss further :)

weedge: yep~  just keep the c++17 change

jan-wassenberg (approved):

jan-wassenberg:
We have an issue with the copybara import mechanism. Working on it.

jan-wassenberg:
Hi @weedge, sorry this has taken so long. Would you mind rebasing?

PR #120, opened 2024-03-28
Fix compilation error when `HWY_COMPILER_GCC_ACTUAL < 1300`

ufownl:
This issue is caused by `hwy::ConvertScalarTo` called in `EmbeddingScaling`. The comment on `EmbeddingScaling` said that it's for rounding the float to bf16 to match Gemma's Embedder. It calls this version of the template function:

```C++
template <typename TTo, typename TFrom, HWY_IF_BF16(TFrom),
          HWY_IF_NOT_SPECIAL_FLOAT(TTo)>
HWY_API HWY_BF16_CONSTEXPR TTo ConvertScalarTo(TFrom in) {
  return static_cast<TTo>(F32FromBF16(in));
}
```

When `HWY_COMPILER_GCC_ACTUAL < 1300` and there is no built-in function `__builtin_bit_cast`, It will not be `constexpr`. However, `constexpr` attribute of `EmbeddingScaling` only follows `Sqrt`, which may cause compilation errors. So I open this PR to fix it.

### DIFF
--- a/gemma.cc
+++ b/gemma.cc
@@ -464,8 +464,16 @@ HWY_NOINLINE void FFW(Activations<TConfig, kBatchSize>& activations,
       activations.ffw_out.data() + batch_idx * kModelDim, pool);
 }
 
+// `EmbeddingScaling` can be constexpr only if `Sqrt` and `hwy::ConvertScalarTo`
+// are both constexpr
+#if HWY_COMPILER_GCC_ACTUAL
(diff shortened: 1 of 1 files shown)


austinvhuang:
Thanks, maybe related -
https://github.com/google/gemma.cpp/pull/119#discussion_r1541089132

jan-wassenberg (approved):
Nice solution, thank you for fixing this :)

PR #122, opened 2024-03-31
.bazelversion: Bazel 7.1.1

LINKIWI:
This change adds a `.bazelversion` file to support per-Bazel-version-reproducible builds via [Bazelisk](https://github.com/bazelbuild/bazelisk).

A stable version of Bazel across builds will also unblock the ability to check in `MODULE.lock`.


jan-wassenberg (approved):
Nice, thank you.

PR #125, opened 2024-04-03
Add standalone tool to compress weights.

szabadka:


gemma.cc
```diff
@@ -134,10 +137,11 @@ hwy::AlignedUniquePtr<Weights<TConfig>> LoadWeights(const Path& checkpoint) {
                      sizeof(layer_view->pre_attention_norm_scale), 1, fptr);
     ok &= 1 == fread(&layer_view->pre_ffw_norm_scale,
                      sizeof(layer_view->pre_ffw_norm_scale), 1, fptr);
+    total_size += sizeof(*layer_view);
   }
   if (!ok) {
-    HWY_ABORT("Failed to read from %s - might be a directory, or too small?",
-              checkpoint.path.c_str());
+    HWY_ABORT("Failed to read from %s - might be a directory, or too small? "
+              "expected size: %zu", checkpoint.path.c_str(), total_size);
```
jan-wassenberg, line 144: When the var is u64, maybe >> 10, then %d and static_cast<unsigned>() to avoid the ugly PRIu64?

util/args.h
```diff
@@ -25,27 +25,43 @@
 
 #include "hwy/base.h"  // HWY_ABORT
 
+#if defined(_WIN32)
+#include <io.h>
+#define F_OK 0
+#define access _access
+#else
+#include "unistd.h"
```
jan-wassenberg, line 33: <unistd.h> please :)

gemma.cc
```diff
@@ -813,6 +817,47 @@ hwy::AlignedFreeUniquePtr<uint8_t[]> GetCompressedWeightsT(
...
+  const hwy::AlignedUniquePtr<Weights<TConfig>> weights =
+      LoadWeights<TConfig>(weights_path);
+  Compressor compressor(pool);
+  ForEachTensor<TConfig>(weights.get(), *c_weights, compressor);
+  compressor.WriteAll(pool, compressed_weights_path.path.c_str());
+
+  c_weights->c_layer_ptrs.~CompressedLayerPointers<TConfig>();
+}
+
+void CompressWeightsT(gcpp::Model model, const Path& weights,
```
jan-wassenberg, line 846: Right above this we have a GetCompressedWeightsT, can't we use that?

austinvhuang:
Very cool I was thinking that it would be nice to have this separated out as its own tool.

This might be a good candidate code to have in the `experimental/` directory. Once it's been tested + refined we could incorporate docs into the top level README.md as the canonical workflow.

CMakeLists.txt
```diff
@@ -115,3 +115,8 @@ foreach (TESTFILE IN LISTS GEMMA_TEST_FILES)
   gtest_discover_tests(${TESTNAME})
 endforeach ()
 endif()  # GEMMA_ENABLE_TESTS
+
+## Tools
+
+add_executable(compress_weights compress_weights.cc)
```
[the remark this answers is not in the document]
szabadka: Done.

compress_weights.cc
```diff
@@ -0,0 +1,131 @@
...
+  std::cerr
+      << "Usage:\n./compress_weights --weights <path to uncompressed weights> "
+      " --model <model type> --compressed_weights <output path>\n";
+  std::cerr << "\n*Arguments*\n\n";
+  args.Help();
+  std::cerr << "\n";
+}
+
+void Run(Args& args) {
+  hwy::ThreadPool pool(0);
```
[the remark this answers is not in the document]
szabadka: I added --num_threads flag.

gemma.cc
```diff
@@ -116,10 +116,13 @@ hwy::AlignedUniquePtr<Weights<TConfig>> LoadWeights(const Path& checkpoint) {
               checkpoint.path.c_str());
   }
   bool ok = true;
+  size_t total_size = 0;
```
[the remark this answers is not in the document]
szabadka: Done.

util/args.h
szabadka: Done.

gemma.cc
szabadka: I wanted to add a separate function, because I think the functionality to fall back to uncompressed weights should be removed from GetCompressedWeights, and the --weights flag should be deprecated in gemma and work only in this separate tool. This should not be a problem since I could not find any working uncompressed weights file anywhere.

jan-wassenberg: Ah, I see. Yes, agree we can remove the --weights soon. It's possibly useful still during the short term for Phil's export script, which does produce uncompressed weights.

jan-wassenberg (approved):
Thanks for making the change!

PR #129, opened 2024-04-05
Add more ops: Sigmoid, (Two)MatVecAdd. Faster TwoMatVec.

veluca93:
drive-by: some build system simplifications

### DIFF
--- a/ops.h
+++ b/ops.h
@@ -88,45 +88,80 @@ template <size_t kOuter>
 HWY_INLINE constexpr size_t RowsPerStrip() {
   // Aim for 128 work items to reduce pool overhead. Must be at least one
   // vector; prefer a power of two for faster division.
+  constexpr size_t kLanes = hn::ScalableTag<float>().MaxLanes();
   constexpr size_t kRowsPerStrip =
-      HWY_MAX(hn::ScalableTag<float>().MaxLanes(),
-              1ULL << hwy::FloorLog2(kOuter / 128));
+      kOuter < 128 ? kLanes
+                   : HWY_MAX(kLanes, 1ULL << hwy::FloorLog2(kOuter / 128));
   return kRowsPerStrip;
 }
 
 // Simple version without tiling nor threading.
-template <size_t kOuter, size_t kInner, typename MatT, size_t kCapacity,
-          typename VecT>
-HWY_INLINE void MatVecLoop(const CompressedArray<MatT, kCapacity>& mat,
-                           const size_t mat_ofs,
-                           const VecT* HWY_RESTRICT vec_aligned,
-                           float* HWY_RESTRICT out) {
-  PROFILER_ZONE("MatVecLoop");
+template <bool kAdd, size_t kOuter, size_t kInner, typename ArrayT,
+          typename VecT, typename AddT>
+HWY_INLINE void MatVecAddLoop(const ArrayT& mat, const size_t mat_ofs,
+                              const VecT* HWY_RESTRICT vec_aligned,
+                              const AddT* HWY_RESTRICT add,
+                              float* HWY_RESTRICT out) {
+  PROFILER_ZONE("MatVecAddLoop");
   const hn::ScalableTag<float> df;
 
   for (size_t idx_row = 0; idx_row < kOuter; ++idx_row) {
     const size_t row_ofs = mat_ofs + idx_row * kInner;
-    out[idx_row] = Dot(df, mat, row_ofs, vec_aligned, kInner);
+    if constexpr (kAdd) {
+      out[idx_row] = hwy::ConvertScalarTo<float>(add[idx_row]) +
+                     Dot(df, mat, row_ofs, vec_aligned, kInner);
+    } else {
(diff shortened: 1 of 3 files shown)


ops.h
```diff
@@ -191,15 +238,15 @@ HWY_INLINE void FullDotProductsForStrip(
...
-// Stores dot products of rows with `vec_aligned` to a buffer, then stores them
-// to `out`.
-template <size_t kOuter, size_t kInner, typename MatT, size_t kCapacity,
-          typename VecT>
-HWY_INLINE void MatVec(const CompressedArray<MatT, kCapacity>& mat,
-                       const size_t mat_ofs,
-                       const VecT* HWY_RESTRICT const vec_aligned,
-                       float* HWY_RESTRICT out, hwy::ThreadPool& pool) {
-  PROFILER_ZONE("MatVec");
+// Stores dot products of rows with `vec_aligned` + add to a buffer (if kAdd),
```
jan-wassenberg, line 241: If we have other changes, consider changing 'add to' to 'add the values from `add`'.

ops.h
```diff
@@ -277,69 +333,105 @@ static HWY_NOINLINE HWY_MAYBE_UNUSED void GeluMulToBF16(
   }
 }
 
+template <class D, HWY_IF_F32_D(D)>
+static HWY_INLINE hn::Vec<D> Sigmoid(D d, hn::Vec<D> v) {
+  using VF = hn::Vec<D>;
+  // Chebyshev polynomial coefficients for rational approximation
+  const VF c0 = Set(d, 0.00949107017368078f);
```
jan-wassenberg, line 340: Let's add hn:: to all ops, otherwise it will break on SVE (and some people are indeed targeting that).

ops.h
```diff
@@ -277,69 +333,105 @@ static HWY_NOINLINE HWY_MAYBE_UNUSED void GeluMulToBF16(
...
+  const VF invtwelve = Set(d, 1.0f / 12.0f);
+  const VF lo = Set(d, -11.5f);
+  const VF hi = Set(d, 11.5f);
+
+  VF f = Clamp(v, lo, hi);
+  f = Mul(f, invtwelve);
+  VF f2 = Add(f, f);
+
+  VF a1 = MulAdd(f2, c0, c1);
+  VF a2 = Sub(MulAdd(f2, a1, c2), c0);
```
jan-wassenberg, line 366: Is this correct? Looks like we could just pre-bake c2-c0?

ops.h
```diff
@@ -277,69 +333,105 @@ static HWY_NOINLINE HWY_MAYBE_UNUSED void GeluMulToBF16(
...
+  VF a3 = Sub(MulAdd(f2, a2, c3), a1);
+  VF a4 = Sub(MulAdd(f2, a3, c4), a2);
+  VF f0 = Sub(MulAdd(f, a4, c5), a3);
+
+  VF b1 = MulAdd(f2, d0, d1);
+  VF b2 = Sub(MulAdd(f2, b1, d2), d0);
```
jan-wassenberg, line 372: Here also.

ops.h
```diff
@@ -277,69 +333,105 @@ static HWY_NOINLINE HWY_MAYBE_UNUSED void GeluMulToBF16(
...
+  VF b3 = Sub(MulAdd(f2, b2, d3), b1);
+  VF f1 = Sub(MulAdd(f, b3, d4), b2);
+
+  return Div(f0, f1);
```
jan-wassenberg, line 376: Should we guard against div by zero?

ops_test.cc
```diff
@@ -356,6 +356,155 @@ void TestAllCreateDistribution() {
   TestCreateDistribution<5000>();
 }
 
+template <size_t kOuter, size_t kInner>
+CompressedArray<float, kOuter * kInner> GenerateMat(size_t offset) {
+  hwy::ThreadPool pool(1);
```
jan-wassenberg, line 361: Should be 0 (main thread only) or probably a few more (3..4?)

ops_test.cc
```diff
@@ -356,6 +356,155 @@ void TestAllCreateDistribution() {
...
+  }
+  return out;
+}
+
+template <size_t length>
+void AssertClose(const hwy::AlignedFreeUniquePtr<float[]>& a,
+                 const hwy::AlignedFreeUniquePtr<float[]>& b) {
+  for (size_t idx = 0; idx < length; idx++) {
+    const float rel_abs_delta =
+        std::abs(a[idx] - b[idx]) / std::max(a[idx], b[idx]);
```
jan-wassenberg, line 408: Also abs of the max? What if both are negative?

jan-wassenberg (changes requested):

ops.h
veluca93: No, f1 is >= 0.05 over the real line.

ops_test.cc
veluca93: Thanks for catching that!

jan-wassenberg (approved):

PR #130, opened 2024-04-05
Improve weight handling.

veluca93:
- Allow scaling of SFP weights
- Allow using uncompressed weights
- Do not try to compress weights in the main model calls
- Reduce code duplication in weight handling with some macros

### DIFF
--- a/gemma.cc
+++ b/gemma.cc
@@ -19,18 +19,18 @@
 // which we pass the filename via macro 'argument'.
 #undef HWY_TARGET_INCLUDE
 #define HWY_TARGET_INCLUDE "gemma.cc"  // NOLINT
-#include "hwy/foreach_target.h"  // IWYU pragma: keep
+#include "hwy/foreach_target.h"        // IWYU pragma: keep
 // Must come after foreach_target.h to avoid redefinition errors.
 // copybara:import_next_line:gemma_cpp
 #include "compression/compress-inl.h"
 // copybara:import_next_line:gemma_cpp
 #include "ops.h"
-// copybara:import_next_line:gemma_cpp
-#include "util/args.h"  // Path
 #include "hwy/contrib/matvec/matvec-inl.h"
 #include "hwy/highway.h"
 #include "hwy/profiler.h"
 #include "hwy/timer.h"
+// copybara:import_next_line:gemma_cpp
+#include "util/args.h"  // Path
 
 // Non-SIMD includes and types. Note that HWY_ONCE is only true on the last
 // compile pass, whereas we want this defined in the first.
@@ -64,6 +64,12 @@
 // copybara:import_next_line:sentencepiece
 #include "src/sentencepiece_processor.h"
 
+// Setting this to true disables fread() calls that read the model file.
+constexpr bool kDryRunFread = false;
+
+// Setting this to false will load and use uncompressed weights.
+constexpr bool kWeightsAreCompressed = true;
+
 namespace gcpp {
 
 template <class TConfig>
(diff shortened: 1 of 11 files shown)


gemma.cc
```diff
@@ -609,9 +697,14 @@ void GenerateImpl(GemmaImpl<TConfig>& gemma, size_t max_tokens,
   Activations<TConfig, 1>& activations = *gemma.state.get();
   Activations<TConfig, kPrefillBatchSize>& prefill_activations =
       *gemma.prefill.get();
-  const CompressedWeights<TConfig>& c_weights =
-      *reinterpret_cast<CompressedWeights<TConfig>*>(
-          gemma.compressed_weights.get());
+
+#if WEIGHTS_ARE_COMPRESSED
```
jan-wassenberg, line 701: How about adding near WEIGHTS_ARE_COMPRESSED a template typedef, something like
using <typename TConfig> WeightsT which is either CompressedWeights or Weights. Then here we do not require an #if.

gemma.cc
```diff
@@ -743,52 +835,37 @@ void ForEachTensor(const Weights<TConfig>* weights,
...
-    func(name, layer ? layer->attn_vec_einsum_w.data() : nullptr,
-         c_layer->c_attn_vec_einsum_w);
-
-    snprintf(name, sizeof(name), "pre_att_ns_%d", layer_idx);
-    func(name, layer ? layer->pre_attention_norm_scale.data() : nullptr,
-         c_layer->c_pre_attention_norm_scale);
+    const Layer<TConfig>* layer = weights ? weights->GetLayer(idx) : nullptr;
+    CompressedLayer<TConfig>* layer_weights = c_weights.GetLayer(idx);
+
+#define CALL_FUNC(memb, n)                          \
```
jan-wassenberg, line 848: Let's have longer names, e.g. identifier/member and key_str or similar. We could rename `name[16]` to name_buf. Putting the string as the first arg might also make this easier to visually scan, because they are about the same length.

gemma.cc
```diff
@@ -743,52 +835,37 @@ void ForEachTensor(const Weights<TConfig>* weights,
...
+    CALL_FUNC(pre_attention_norm_scale, "pre_att_ns");
+#undef CALL_FUNC
   }
 }
 
 template <class TConfig>
-hwy::AlignedFreeUniquePtr<uint8_t[]> GetCompressedWeights(
-    const Path& weights_path, const Path& cache, hwy::ThreadPool& pool) {
+hwy::AlignedFreeUniquePtr<uint8_t[]> LoadCompressedWeights(
+    const Path& weights_path, hwy::ThreadPool& pool) {
```
jan-wassenberg, line 864: Remove _path to match caller?

jan-wassenberg (changes requested):

gemma.cc
```diff
@@ -88,89 +94,165 @@ struct Layer {
...
+    } else {
+      fprintf(stderr, "Loading Parameters (layer=%zu, size %zu): %s\n", layer,
+              size, name);
+    }
+    if constexpr (!kDryRunFread) {
+      ok &= 1 == fread(var, size, 1, fptr);
+      total_size += size;
+    }
+  };
+  do_fread(&(weights->embedder_input_embedding), -1, "embedder_input_embedding",
```
jan-wassenberg, line 190: -1 for size_t will at least raise warnings. Let's change do_fread arg to int, the layer=%zu to %d.

jan-wassenberg (changes requested):
Just one more for do_fread :)

gemma.cc
veluca93: Done.

jan-wassenberg (approved):
Nice!

util/app.h
```diff
@@ -37,8 +37,8 @@
 // copybara:import_next_line:gemma_cpp
 #include "gemma.h"
 // copybara:import_next_line:gemma_cpp
```
jan-wassenberg, line 39: The next_line comment must also be removed whenever we remove/reorder an include of the files in gemma_cpp.
FYI this will no longer be necessary when source files have been moved to gemma/, as is planned.

gemma.cc
```diff
@@ -19,18 +19,18 @@
...
 #undef HWY_TARGET_INCLUDE
 #define HWY_TARGET_INCLUDE "gemma.cc"  // NOLINT
-#include "hwy/foreach_target.h"  // IWYU pragma: keep
+#include "hwy/foreach_target.h"        // IWYU pragma: keep
 // Must come after foreach_target.h to avoid redefinition errors.
 // copybara:import_next_line:gemma_cpp
 #include "compression/compress-inl.h"
 // copybara:import_next_line:gemma_cpp
 #include "ops.h"
 // copybara:import_next_line:gemma_cpp
```
jan-wassenberg, line 28: Please move this next_line comment in front of the new location of the util/args include.

jan-wassenberg (approved):

PR #131, opened 2024-04-06
Add a benchmark and additional tests.

veluca93:
Also add a script to help running sanitizer builds, and do some cleanup.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,11 @@
 name: build
 
 # Trigger on push, pull request, or via manual dispatch.
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+  pull_request:
+    types: [opened, reopened, labeled, unlabeled, synchronize]
+  workflow_dispatch:
 
 jobs:
   build:
--- a/benchmark.cc
+++ b/benchmark.cc
@@ -1,0 +1,293 @@
+#include <algorithm>
+#include <fstream>
+#include <iostream>
+#include <ostream>
+#include <sstream>
+#include <vector>
+
+#include "hwy/base.h"
+#include "hwy/contrib/thread_pool/thread_pool.h"
+#include "hwy/highway.h"
+#include "hwy/per_target.h"
+#include "hwy/profiler.h"
+#include "hwy/timer.h"
+#include "nlohmann/json.hpp"
+// copybara:import_next_line:gemma_cpp
+#include "util/app.h"
+// copybara:import_next_line:gemma_cpp
+#include "util/args.h"
+
+using json = nlohmann::json;
+
+class BenchmarkArgs : public gcpp::ArgsBase<BenchmarkArgs> {
+ public:
+  BenchmarkArgs(int argc, char* argv[]) { InitAndParse(argc, argv); }
+
+  gcpp::Path goldens;
+  gcpp::Path summarize_text;
+  gcpp::Path cross_entropy;
+  gcpp::Path trivia_qa;
+  size_t max_questions;
+  size_t batch_tokens;
+
+  template <class Visitor>
+  void ForEach(const Visitor& visitor) {
+    visitor(goldens.path, "goldens_dir", std::string(""),
+            "Directory containing golden files", 2);
+    visitor(summarize_text.path, "summarize_text", std::string(""),
+            "Path to text file to summarize", 2);
+    visitor(cross_entropy.path, "cross_entropy", std::string(""),
+            "Path to text file to compute cross entropy on", 2);
+    visitor(trivia_qa.path, "trivia_qa", std::string(""),
+            "Path to json file containing TriviaQA entries", 2);
+    visitor(max_questions, "max_questions", (size_t)20,
+            "Maximum number of questions to ask from --trivial_qa input", 2);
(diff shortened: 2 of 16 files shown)


.github/workflows/build.yml
```diff
@@ -72,4 +76,4 @@ jobs:
         with:
           path: ~/.cache/bazel
           key: bazel-${{ runner.os }}
-      - run: bazel build --cxxopt=-std=c++20 //...
+      - run: bazel build -c opt --cxxopt=-std=c++20 //...
```
jan-wassenberg, line 79: FYI we deliberately omitted opt to speed up builds. Internal CI has a ~7 min deadline and we are pretty close to exceeding that :/ Bazel is slower than CMake. Any thoughts on how else to speed up the build?

benchmark.cc
```diff
@@ -0,0 +1,293 @@
...
+#include <vector>
+
+#include "hwy/base.h"
+#include "hwy/contrib/thread_pool/thread_pool.h"
+#include "hwy/highway.h"
+#include "hwy/per_target.h"
+#include "hwy/profiler.h"
+#include "hwy/timer.h"
+#include "nlohmann/json.hpp"
+#include "util/app.h"
```
jan-wassenberg, line 16: These two require the next_line comment.

benchmark.cc
```diff
@@ -0,0 +1,293 @@
...
+#include "util/args.h"
+
+using json = nlohmann::json;
+
+class BenchmarkArgs : public gcpp::ArgsBase<BenchmarkArgs> {
+ public:
+  BenchmarkArgs(int argc, char* argv[]) { InitAndParse(argc, argv); }
+
+  std::string goldens_dir;
```
jan-wassenberg, line 26: Consider using Path for these?

benchmark.cc
```diff
@@ -0,0 +1,293 @@
...
+
+void LogSpeedStats(const double time_start, size_t total_tokens) {
+  const double time_end = hwy::platform::Now();
+  const double time_elapsed = time_end - time_start;
+  const double tok_sec = total_tokens / time_elapsed;
+  std::cout << total_tokens << " tokens in " << time_elapsed << " seconds"
+            << " [" << tok_sec << " tokens / sec" << "]\n";
+}
+
+std::pair<std::string, int> query_model(
```
jan-wassenberg, line 60: PascalCase naming please :)

benchmark.cc
```diff
@@ -0,0 +1,293 @@
...
+    return true;
+  };
+  if (app.verbosity >= 2) {
+    std::cout << args.max_tokens << " " << args.max_generated_tokens << " "
+              << args.temperature;
+  }
+  GenerateGemma(model, args.max_tokens, args.max_generated_tokens,
+                args.temperature, prompt, /*abs_pos=*/0, kv_cache, pool,
+                inner_pool, stream_token, accept_token, gen, app.verbosity);
+  // const double time_end = hwy::platform::Now();
```
jan-wassenberg, line 95: Remove? This seems to be in LogSpeedStats.

benchmark.cc
```diff
@@ -0,0 +1,295 @@
...
+      std::cout << "WRONG\n\n";
+    }
+    if (++i >= max_questions) break;
+  }
+  printf("Correct answers: %zu / %zu\n", correct_answers, i);
+  return EXIT_SUCCESS;
+}
+
+/* Run this in the same way as gemma, p.ex.:
+ ./benchmark --tokenizer tokenizer.spm --model 2b-it --compressed_weights \
```
jan-wassenberg, line 253: --weights is now preferred, right?

gemma.cc
```diff
@@ -1090,5 +1235,16 @@ void CompressWeights(gcpp::Model model, const Path& weights,
   (model, weights, compressed_weights, pool);
 }
 
+float ComputeCrossEntropy(Gemma& gemma, size_t max_tokens,
+                          const std::vector<int>& prompt, KVCache& kv_cache,
+                          hwy::ThreadPool& pool, hwy::ThreadPool& inner_pool,
+                          int verbosity) {
+  pool.SetWaitMode(hwy::PoolWaitMode::kSpin);
```
jan-wassenberg, line 1242: Is this time-critical? We can stick with the default/current wait mode, I think.

gemma_test.cc
```diff
@@ -0,0 +1,303 @@
...
+// limitations under the License.
+
+// copybara:import_next_line:gemma_cpp
+#include "gemma.h"
+
+#include <thread>
+
+#include "gtest/gtest.h"
+#include "hwy/contrib/thread_pool/thread_pool.h"
+#include "ops.h"
```
jan-wassenberg, line 23: Also requires next_line.

jan-wassenberg (changes requested):

.github/workflows/build.yml
veluca93: No strong opinions :-)

jan-wassenberg: Let's remove opt then for now.

gemma.cc
veluca93: It's part of a benchmark, which I assume why this is here :-)

jan-wassenberg (approved):

PR #137, opened 2024-04-09
Clarified README

zond:
Made it more visible that the recurrent weights are at a different Kaggle page.


jan-wassenberg (approved):
Thanks!

PR #142, opened 2024-04-10
Refactor data structures to reduce memory usage

ufownl:
This PR moves the Gemma attention weights and Griffin weights into a union to share memory and reduces the additional overhead when not using the Griffin model.

### DIFF
--- a/gemma/gemma.cc
+++ b/gemma/gemma.cc
@@ -84,32 +84,45 @@ struct Layer {
       (kHeads + 2 * kKVHeads) * kQKVDim * kModelDim;
   // 2x for (gelu gating vector, gated vector)
   static constexpr size_t kGatingEinsumWSize = 2 * kFFHiddenDim * kModelDim;
+  static constexpr size_t kConv1dWidth = TConfig::kConv1dWidth;
+  static constexpr bool kFFBiases = TConfig::kFFBiases;
+  static constexpr size_t kAOBiaseDim =
+      TConfig::kSoftmaxAttnOutputBiases ? kModelDim : 0;
+  static constexpr size_t kGriffinDim =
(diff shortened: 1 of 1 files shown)


gemma/gemma.cc
```diff
@@ -84,32 +84,45 @@ struct Layer {
       (kHeads + 2 * kKVHeads) * kQKVDim * kModelDim;
   // 2x for (gelu gating vector, gated vector)
   static constexpr size_t kGatingEinsumWSize = 2 * kFFHiddenDim * kModelDim;
+  static constexpr size_t kConv1dWidth = TConfig::kConv1dWidth;
+  static constexpr bool kFFBiases = TConfig::kFFBiases;
+  static constexpr size_t kAOBiaseDim =
```
jan-wassenberg, line 89: Typo (extra e), maybe instead kAttnBiasDim?

ufownl: Oops, my bad.

gemma/gemma.cc
```diff
@@ -84,32 +84,45 @@ struct Layer {
...
+
+  union {
+    struct {
+      ArrayT<float, kAttVecEinsumWSize> attn_vec_einsum_w;
+      ArrayT<float, kQKVEinsumWSize> qkv_einsum_w;
+      ArrayT<float, kAOBiaseDim> attention_output_biases;
+    };
+
+    struct {
+      ArrayT<float, kGriffinDim * kGriffinDim> griffin_linear_x_w;
```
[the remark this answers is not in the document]
ufownl: I agree, your solution should be better.

jan-wassenberg (approved):
Nice, thanks for making the change :)

PR #144, opened 2024-04-12
Added functionality of storing layers activations output.

atorero:
Adding a way to save activations output after applying each layer. This can be useful for two purposes:

- Debugging the model and comparing the output to another working implementation
- Quantising weights or activations and getting insights of the accumulated error.

There will be a similar functionality in the Python code and a diffing tool that will compare two outputs and calculating some metrics for them. See the output of such a tool when comparing outputs of (3 tokens)x(27 layers) between the Python and the C++ (8-bit) Recurrent Gemma implementations:

image


BUILD.bazel
```diff
@@ -170,3 +170,22 @@ cc_binary(
...
+        ":gemma_lib",
+        # "//base",
+        "//compression:compress",
+        "@hwy//:hwy",
+        "@hwy//:nanobenchmark",
+        "@hwy//:profiler",
+        "@hwy//:thread_pool",
+        "@nlohmann_json//:json",
+    ],
+)
```
jan-wassenberg, line 191: Generally good to have trailing newlines :)

debug_prompt.cc
```diff
@@ -0,0 +1,135 @@
+
+#include <fstream>
+#include <iostream>
+#include <string>
+
+#include "gemma.h"
+#include "nlohmann/json.hpp"
+// copybara:import_next_line:gemma_cpp
```
jan-wassenberg, line 8: This is no longer necessary for gemma headers, only for sentencepiece. Please remove both here :)

debug_prompt.cc
```diff
@@ -0,0 +1,135 @@
...
+    output_f << json_logger.json_output.dump();
+    if (!output_f) {
+      std::cout << "Writing to file failed" << std::endl;
+      return EXIT_FAILURE;
+    }
+    output_f.close();
+  }
+
+  return EXIT_SUCCESS;
+}
```
jan-wassenberg, line 135: Trailing newline please :)

gemma.cc
```diff
@@ -24,11 +24,11 @@
 // copybara:import_next_line:gemma_cpp
 #include "compression/compress-inl.h"
 // copybara:import_next_line:gemma_cpp
-#include "ops.h"
 #include "hwy/contrib/matvec/matvec-inl.h"
 #include "hwy/highway.h"
 #include "hwy/profiler.h"
 #include "hwy/timer.h"
+#include "ops.h"
 // copybara:import_next_line:gemma_cpp
```
jan-wassenberg, line 32: Please remove all gemma_cpp import_next_line :)

gemma.cc
```diff
@@ -1588,7 +1605,8 @@ void GenerateGemma(Gemma& gemma, RuntimeConfig runtime_config,
   GenerateGemma(
       gemma, runtime_config.max_tokens, runtime_config.max_generated_tokens,
       runtime_config.temperature, prompt, start_pos, kv_cache, pool, inner_pool,
-      stream_token, [](int) { return true; }, gen, runtime_config.verbosity);
+      stream_token, [](int) { return true; }, gen, runtime_config.verbosity,
+      nullptr);
```
jan-wassenberg, line 1609: Comment `/*layers_output=*/` would be helpful here and at other call sites.

jan-wassenberg (changes requested):
Nice, thanks for adding! Some minor suggestions.

BTW I think it's getting more and more useful to more towards an AuxOut struct for Transfomer/Generate, the arg list is getting unmanageable.

BUILD.bazel
atorero: Done :)

debug_prompt.cc
atorero: Done

debug_prompt.cc
atorero: Sure :)

gemma.cc
atorero: Done

gemma.cc
atorero: Done :)

atorero:
> Apologies, I did not notice the pull request is targeting main branch. Please generally target dev :)

Sure, sorry for the confusion :) https://github.com/google/gemma.cpp/pull/145

PR #145, opened 2024-04-12
Added functionality of storing layers activations output.

atorero:
Adding a way to save activations output after applying each layer. This can be useful for two purposes:

- Debugging the model and comparing the output to another working implementation
- Quantising weights or activations and getting insights of the accumulated error.

There will be a similar functionality in the Python code and a diffing tool that will compare two outputs and calculating some metrics for them. See the output of such a tool when comparing outputs of (3 tokens)x(27 layers) between the Python and the C++ (8-bit) Recurrent Gemma implementations:

image


jan-wassenberg (approved):
Nice, thanks for updating the PR!

PR #151, opened 2024-04-20
Make stream_token can break prefill, too 

zeerd:
Re-Open #94 because of the conflict.

PR #156, opened 2024-04-23
Use lambda to split function and Make stream_token can break prefill

zeerd:
Re-Open #94 and #151 for dev branch and conflict.

```
When we give the gemma a large input(ex. a file), we need a way to stop it.
```


jan-wassenberg (approved):
Thank you for re-opening!

PR #166, opened 2024-04-29
Factor out deinterleaving of bf16 vectors for MatVecs.

samkaufman:
This specializes bf16-f32 and bf16-bf16 vector-matrix multiplications to first convert bf16 vectors into f32 buffers of vector-length strips of even- and odd-indexed values.

The 2B, bf16 model running on my Zen 1 machine sees ~10% throughput improvements to single-threaded prefill, single-threaded generation, and multi-threaded prefill, but only a marginal improvement to multi-threaded generation throughput.

This PR does not implement support for SFP.

## TODOs:
- [x] Check for native BF16 support.
- [x] Lift out allocations of f32 vectors. (@jan-wassenberg volunteered to handle this.)
- [x] Clean up some MatVecAdd code duplication.

### DIFF
--- a/compression/compress-inl.h
+++ b/compression/compress-inl.h
@@ -51,20 +51,6 @@ namespace gcpp {
 namespace HWY_NAMESPACE {
 namespace hn = hwy::HWY_NAMESPACE;
 
-template <class DF, HWY_IF_F32_D(DF)>
-HWY_INLINE void Bf16ToF32EO(const DF df32,
-                            const hwy::bfloat16_t* HWY_RESTRICT in,
-                            hn::Vec<DF>& v_even,
-                            hn::Vec<DF>& v_odd) {
-  const hn::Repartition<hwy::bfloat16_t, DF> dbf16;
-  const hn::RebindToUnsigned<decltype(df32)> du32;
-
-  const auto odd = Set(du32, 0xFFFF0000u);
-  const auto interleaved = BitCast(du32, LoadU(dbf16, in));
-  v_even = BitCast(df32, hn::ShiftLeft<16>(interleaved));
-  v_odd = BitCast(df32, And(interleaved, odd));
-}
-
 // Enables generic code independent of compression type.
 template <typename T>  // primary, must specialize
 struct CompressTraits {};
@@ -263,19 +249,19 @@ struct CompressTraits<hwy::bfloat16_t> {
 
     VF32 be0, bo0, be1, bo1;
     for (size_t i = 0; i < num; /* i += 2 * N */) {
+      const auto interleaved0 = hn::LoadU(dbf16, in + in_ofs + i);
       const VF32 ae0 = Load(df32, vec_aligned + i);
       const VF32 ao0 = Load(df32, vec_aligned + i + (N / 2));
-      Bf16ToF32EO(df32, in + in_ofs + i, be0, bo0);
+      sum0 = hn::MulAdd(ae0, hn::PromoteEvenTo(df32, interleaved0), sum0);
+      sum1 = hn::MulAdd(ao0, hn::PromoteOddTo(df32, interleaved0), sum1);
       i += N;
-      sum0 = hn::MulAdd(ae0, be0, sum0);
-      sum1 = hn::MulAdd(ao0, bo0, sum1);
 
+      const auto interleaved1 = hn::LoadU(dbf16, in + in_ofs + i);
       const VF32 ae1 = Load(df32, vec_aligned + i);
       const VF32 ao1 = Load(df32, vec_aligned + i + (N / 2));
-      Bf16ToF32EO(df32, in + in_ofs + i, be1, bo1);
+      sum2 = hn::MulAdd(ae1, hn::PromoteEvenTo(df32, interleaved1), sum2);
+      sum3 = hn::MulAdd(ao1, hn::PromoteOddTo(df32, interleaved1), sum3);
       i += N;
-      sum2 = hn::MulAdd(ae1, be1, sum2);
-      sum3 = hn::MulAdd(ao1, bo1, sum3);
     }
 
     sum0 = Add(sum0, sum1);
--- a/compression/compress-inl.h
+++ b/compression/compress-inl.h
@@ -72,7 +72,7 @@ struct CompressTraits {};
 template <>
 struct CompressTraits<float> {
   using MatT = float;
-  static constexpr bool supports_eo = false;
+  static constexpr bool kSupportsEvenOdd = false;
 
   template <class DF, HWY_IF_F32_D(DF)>
   static HWY_INLINE void Compress(DF df, const float* HWY_RESTRICT in,
@@ -126,7 +126,7 @@ struct CompressTraits<float> {
 template <>
 struct CompressTraits<hwy::bfloat16_t> {
   using MatT = hwy::bfloat16_t;
-  static constexpr bool supports_eo = true;
+  static constexpr bool kSupportsEvenOdd = true;
 
   template <class DF, HWY_IF_F32_D(DF)>
   static HWY_INLINE void Compress(DF df, const float* HWY_RESTRICT in,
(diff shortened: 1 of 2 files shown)


compression/compress-inl.h
```diff
@@ -51,13 +51,28 @@ namespace gcpp {
...
+HWY_INLINE void Bf16ToF32EO(const DF df32,
+                            const hwy::bfloat16_t* HWY_RESTRICT in,
+                            hn::Vec<DF>& v_even,
+                            hn::Vec<DF>& v_odd) {
+  const hn::Repartition<hwy::bfloat16_t, DF> dbf16;
+  const hn::RebindToUnsigned<decltype(df32)> du32;
+
+  const auto odd = Set(du32, 0xFFFF0000u);
+  const auto interleaved = BitCast(du32, LoadU(dbf16, in));
+  v_even = BitCast(df32, hn::ShiftLeft<16>(interleaved));
```
jan-wassenberg, line 64: We now have PromoteEvenTo and PromoteOddTo ops you could use :)

compression/compress-inl.h
```diff
@@ -51,13 +51,28 @@ namespace gcpp {
...
+}
+
 // Enables generic code independent of compression type.
 template <typename T>  // primary, must specialize
 struct CompressTraits {};
 
 template <>
 struct CompressTraits<float> {
   using MatT = float;
+  static constexpr bool supports_eo = false;
```
jan-wassenberg, line 75: Please use kUppercase naming. Maybe also write out to EvenOdd?

gemma/ops.h
```diff
@@ -92,6 +92,43 @@ HWY_INLINE constexpr size_t RowsPerStrip() {
...
+
+  VF32 veven, vodd;
+  for (size_t i = 0; i < size; i += hn::Lanes(dbf16)) {
+    Bf16ToF32EO(df, vec_aligned + i, veven, vodd);
+    hn::Store(veven, df, out + i);
+    hn::Store(vodd, df, out + i + hn::Lanes(df));
+  }
+}
+
+HWY_INLINE void ToEvenOddF32(
```
jan-wassenberg, line 115: Is it possible to avoid calling this for f32 inputs? Seems like we'd be better off doing a direct f32*f32 instead.

gemma/ops.h
```diff
@@ -113,12 +150,38 @@ HWY_INLINE void MatVecAddLoop(const ArrayT& mat, const size_t mat_ofs,
...
+    const CompressedArray<hwy::bfloat16_t, kCapacity>& mat,
+    const size_t mat_ofs,
+    const VecT* HWY_RESTRICT vec_aligned,
+    const AddT* HWY_RESTRICT add,
+    float* HWY_RESTRICT out) {
+  PROFILER_ZONE("MatVecAddLoop");
+
+  const hn::ScalableTag<float> df;
+
+  const auto vec_dequant = hwy::AllocateAligned<float>(kInner);
```
jan-wassenberg, line 165: Allocation can be quite slow, let's move this into gemma.cc's Activations. That would require plumbing through an extra tmp arg here, and the std::array storage should probably be the largest per-call size * max number of threads (say 128 or 256). Would you prefer if I made this change?

gemma/ops.h
```diff
@@ -113,12 +150,38 @@ HWY_INLINE void MatVecAddLoop(const ArrayT& mat, const size_t mat_ofs,
...
+}
+
 template <size_t kOuter, size_t kInner, typename ArrayT, typename VecT>
 HWY_INLINE void MatVecLoop(const ArrayT& mat, const size_t mat_ofs,
                            const VecT* HWY_RESTRICT vec_aligned,
                            float* HWY_RESTRICT out) {
-  MatVecAddLoop<false, kOuter, kInner, ArrayT, VecT, VecT>(
-      mat, mat_ofs, vec_aligned, /*add=*/nullptr, out);
+  MatVecAddLoop<false, kOuter, kInner>(
+      mat, mat_ofs, vec_aligned, /*add=*/(VecT*)nullptr, out);
```
jan-wassenberg, line 184: Nit: static_cast<VecT*> please :)

gemma/ops.h
```diff
@@ -254,27 +321,102 @@ HWY_INLINE void MatVecAdd(const ArrayT& mat, const size_t mat_ofs,
...
+                                                 out + r0);
+  }
+}
+
+// A specialization of MatVecAdd to float32 vectors which first rearranges the
+// vector to even-odd layout.
+template <bool kAdd, size_t kOuter, size_t kInner, typename ArrayT,
+          typename AddT,
+          std::enable_if_t<CompressTraits<typename ArrayT::value_type>::supports_eo, bool> = true>
+HWY_INLINE void MatVecAdd(const ArrayT& mat, const size_t mat_ofs,
```
jan-wassenberg, line 345: Can you help me understand the benefit of this one, or can we just skip it?

jan-wassenberg (changes requested):
Nice, great to see this coming together :) Thanks for sending the PR. Some sugestions:

gemma/ops.h
jan-wassenberg: Ah, I see, makes sense.

jan-wassenberg:
> @jan-wassenberg One more question: what's the best way to check that the target doesn't have a native bf16 product/dot product support (e.g., AVX512_BF16)? You previously pointed me at a [highway PR](https://github.com/google/highway/pull/2112/), but it looks like Copybara scrubbed the branch when the PR was dropped.

We can check `#if defined(HWY_NATIVE_DOT_BF16) && HWY_NATIVE_DOT_BF16` :)

compression/compress-inl.h
```diff
@@ -219,11 +221,60 @@ struct CompressTraits<hwy::bfloat16_t> {
...
+    const hn::RebindToUnsigned<decltype(df32)> du32;
+    using VU32 = hn::VFromD<decltype(du32)>;
+    const VU32 odd = Set(du32, 0xFFFF0000u);
+
+    VF32 be0, bo0, be1, bo1;
+    for (size_t i = 0; i < num; /* i += 2 * N */) {
+      const auto interleaved0 = hn::LoadU(dbf16, in + in_ofs + i);
+      const VF32 ae0 = Load(df32, vec_aligned + i);
+      const VF32 ao0 = Load(df32, vec_aligned + i + (N / 2));
+      sum0 = hn::MulAdd(ae0, hn::PromoteEvenTo(df32, interleaved0), sum0);
```
jan-wassenberg, line 255: FYI we will soon add MulEvenAdd which fuses these two. No need to change now, it will require updating the Highway version.

gemma/ops.h
jan-wassenberg: Got it, thanks. I now understand why this is useful even for already-f32 vectors.

gemma/ops.h
```diff
@@ -254,27 +320,71 @@ HWY_INLINE void MatVecAdd(const ArrayT& mat, const size_t mat_ofs,
...
+                                                 num_rows, vec_aligned, add, 
+                                                 out + r0);
+  }
+}
+
+// A specialization of MatVecAdd to float32 vectors which first rearranges the
+// vector to even-odd layout.
+template <bool kAdd, size_t kOuter, size_t kInner, typename ArrayT,
+          typename VecT, typename AddT,
+          std::enable_if_t<
```
jan-wassenberg, line 343: Consider replacing with `HWY_IF_SAME2(VecT, float, hwy::bfloat16_t)`.

gemma/ops.h
jan-wassenberg: Done in #173, see `even_odd` :)

jan-wassenberg (approved):
Nice, looks good to me, thanks for updating! Can you give it a quick sanity check also with sfp weights (e.g. those prefixed 1.1 on Kaggle) to make sure that also still works?

jan-wassenberg:
Thanks for confirming :D

jan-wassenberg:
Internal CI caught some unused vars:
```
third_party/gemma_cpp/gemma/ops.h:101:14: error: unused variable 'odd' [-Werror,-Wunused-variable]
  101 |   const auto odd = Set(du32, 0xFFFF0000u);
      |              ^~~
third_party/gemma_cpp/gemma/ops.h:364: error: unused variable 'df' [-Werror,-Wunused-variable]
  364 |   const hn::ScalableTag<float> df;
      |                                ^~
third_party/gemma_cpp/gemma/ops.h:366: error: unused variable 'kNumStrips' [-Werror,-Wunused-variable]
  366 |   constexpr size_t kNumStrips = kOuter / kRowsPerStrip;
```
Please fix :)

PR #169, opened 2024-04-30
work with cmake install

xinpingwang:
currently, gemma only works at `build` dir, it better to support `cmake --install` to install the build output to system directory and easy to use `gemma` globally. I made two changes:
1. allow `gemma` and `libgemma` to be install; 
2. change `target_link_libraries` of `sentencepiece` to `sentencepiece-static`, so that the build output is portable

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,15 +78,17 @@ set_property(TARGET libgemma PROPERTY CXX_STANDARD 17)
 set_target_properties(libgemma PROPERTIES PREFIX "")
 set_property(TARGET libgemma PROPERTY POSITION_INDEPENDENT_CODE ON)
 target_include_directories(libgemma PUBLIC ./)
-target_link_libraries(libgemma hwy hwy_contrib sentencepiece)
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Thank you!

xinpingwang:
> Thank you!

Could you help release a new version after this pr merged. so that I can create a new formula for homebrew Homebrew/homebrew-core#166881

jan-wassenberg:
Apologies, I did not see that this PR targets the main branch, thus our import automation failed. Would you please change it to dev? Soon after it lands in dev, we can do a release.

xinpingwang:
> Apologies, I did not see that this PR targets the main branch, thus our import automation failed. Would you please change it to dev? Soon after it lands in dev, we can do a release.

ok, updated.

jan-wassenberg (approved):

jan-wassenberg:
Thanks! Restarted the import.

xinpingwang:
Merged success to dev, could you help release a new version. Thanks.

jan-wassenberg:
Yes, we will do one soon :)

PR #170, opened 2024-04-30
Use more parallelism in the QKV projections in MQA mode.

szabadka:
Instead of MatVecLoop, we use MatVec and we combine k and v into one 2 * kQKVDim long vector so that K and V projections can be combined into one MatVec operation.

Benchmark results (summarization with 1600 tokens for prefill and essay writing with 500 tokens for generation):

```
4                 9.81 t/s     9.96 t/s       8.39 t/s     8.46 t/s
18               31.50 t/s    36.67 t/s      23.10 t/s    25.83 t/s
32               45.36 t/s    58.91 t/s      27.60 t/s    31.25 t/s
64               57.72 t/s    80.64 t/s      35.40 t/s    39.76 t/s
```

### DIFF
--- a/gemma/gemma.cc
+++ b/gemma/gemma.cc
@@ -399,9 +399,9 @@ struct Activations {
   static constexpr size_t kQKVDim = TConfig::kQKVDim;
   static constexpr size_t kHeads = TConfig::kHeads;
   static constexpr size_t kKVHeads = TConfig::kKVHeads;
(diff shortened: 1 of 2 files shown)


jan-wassenberg:
FYI we are working on a fix for this change, it breaks 7B(MHA).

jan-wassenberg:
#172.

PR #176, opened 2024-05-02
Use more parallelism in the QKV projections of the MHA block.

szabadka:
We compute all three projections with one MatVec and then copy the kv part to the cache.

Benchmark results for 7b-it model that uses MHA blocks (summarization with 1600 tokens for prefill and essay writing with 500 tokens for generation):

```
32               13.75 t/s    14.80 t/s       9.22 t/s     9.77 t/s
64               19.89 t/s    24.83 t/s      12.46 t/s    13.66 t/s
```

### DIFF
--- a/gemma/gemma.cc
+++ b/gemma/gemma.cc
@@ -402,10 +402,11 @@ struct Activations {
   static constexpr size_t kCacheLayerSize = kKVHeads * kQKVDim * 2;
   static constexpr size_t kCachePosSize =
       TConfig::kGemmaLayers * kCacheLayerSize;
+  static constexpr size_t kQDim = kHeads == kKVHeads ? kQKVDim * 3 : kQKVDim;
 
   std::array<float, kBatchSize * kModelDim> x;  // input
   std::array<float, kBatchSize * kModelDim> pre_att_rms_out;
-  std::array<float, kBatchSize * kHeads * kQKVDim> q;  // query vector
+  std::array<float, kBatchSize * kHeads * kQDim> q;  // query vector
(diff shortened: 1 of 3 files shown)


gemma/gemma.cc
```diff
@@ -745,34 +745,23 @@ HWY_NOINLINE void Attention(size_t batch_start, size_t batch_idx, size_t layer,
...
     // Multi-Head Attention
+    static_assert(TConfig::kInterleaveQKV);
+
+    float* HWY_RESTRICT qkv =
+        activations.q.data() + batch_idx * kHeads * kQKVDim * 3;
+    MatVec<kHeads * kQKVDim * 3, kModelDim>(
+        layer_weights->qkv_einsum_w, 0, x, activations.even_odd.data(), qkv,
+        pool);
+
     pool.Run(0, kHeads, [&](const uint64_t head, size_t thread) HWY_ATTR {
```
jan-wassenberg, line 756: Now that MatVec is outside the parallel section, I am not sure it is still worthwhile. Want to try a regular loop here?

jan-wassenberg (commented):
Great to get rid of MatVecLoop! Just to confirm that the MQA 2B still works?

szabadka:
> Great to get rid of MatVecLoop! Just to confirm that the MQA 2B still works?

Yes, I tested that it still works.

gemma/gemma.cc
szabadka: The regular loop is slower because Attn() still has the Q*K dot product calculations that get longer and longer. It starts out at roughly the same speed but after 1600 tokens it ends up half as fast.

PR #217, opened 2024-06-07
Add support for custom sampling function to runtime config.

szabadka:
With this addition the ComputeCrossEntropy function can be moved to its own library, because now we can compute it using only the public API functions from gemma.h


jan-wassenberg (approved):
Very nice!

PR #224, opened 2024-06-10
Use CompressedWeights<TConfig<float>> in backpropagation.

szabadka:
kWeightsAreCompressed are removed and LoadRawWeights is moved to compress_weights.cc

### DIFF
--- a/compression/compress.h
+++ b/compression/compress.h
@@ -79,6 +79,9 @@ class CompressedArray {
   MatT* data() { return data_.data(); }
   const MatT* data() const { return data_.data(); }
 
+  MatT& operator[](size_t pos) { return data_[pos]; }
+  const MatT& operator[](size_t pos) const { return data_[pos]; }
+
   float scale() const { return scale_[0]; }
   void set_scale(float scale) { scale_[0] = scale; }
 
--- a/compression/compress.h
+++ b/compression/compress.h
@@ -79,9 +79,6 @@ class CompressedArray {
   MatT* data() { return data_.data(); }
   const MatT* data() const { return data_.data(); }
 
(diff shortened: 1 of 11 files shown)


compression/compress.h
```diff
@@ -79,6 +79,9 @@ class CompressedArray {
   MatT* data() { return data_.data(); }
   const MatT* data() const { return data_.data(); }
 
+  MatT& operator[](size_t pos) { return data_[pos]; }
```
jan-wassenberg, line 82: Are we OK with this not working for MatT != bfloat16_t and float?  Maybe safest to remove, so that users are encouraged to go through compress-inl functions?

gemma/gemma.cc
```diff
@@ -889,8 +889,9 @@ Gemma::Gemma(GemmaTokenizer&& tokenizer, Model model_type, Type weight_type,
       tokenizer_(std::move(tokenizer)),
       model_type_(model_type),
       weight_type_(weight_type) {
-  weights_u8_ =
-      CallForModelAndWeight<AllocateWeightsF>(model_type, weight_type, pool);
+  HWY_ASSERT(weight_type == Type::kF32);
```
jan-wassenberg, line 892: Any reason not to support other types here? We could just CallForModelAndWeight, right?

jan-wassenberg (approved):
Wonderful, thank you for simplifying the weight loader :)

gemma/gemma.cc
jan-wassenberg: Yeah, I understand. Just thought we could move the assert to the training code, because it seems this ctor might be useful for other weight types also.

jan-wassenberg (approved):

PR #266, opened 2024-06-21
Fix KV cache size calculation error

ufownl:
After refactoring the KV cache size calculation, the wrong functor was used to calculate the size of KV cache. It leads to KV cache buffer overflow.


jan-wassenberg (approved):
Sorry I didn't see this during review. Thanks for fixing! I've also checked the other occurrences.

PR #344, opened 2024-08-10
Implement `start_pos` per query for batch interface

ufownl:
Refs #338

I noticed that the original padding code is not used, and the current implementation truncates the prompts to align to the shortest length to avoid accessing out-of-range. So before calling the batch interface, users should insert `<pad>` tokens in the front of the prompts to align them to the maximum length.

### DIFF
--- a/gemma/gemma-inl.h
+++ b/gemma/gemma-inl.h
@@ -233,7 +233,7 @@ class GemmaAttention {
   // Fills activations.q and computes KV. For kIsMHA, a single MatMul suffices
   // and we later copy KV from q to KVCache. Otherwise, a second MatMul writes
   // KV directly to KVCache.
-  HWY_NOINLINE void ComputeQKV(const size_t batch_start,
+  HWY_NOINLINE void ComputeQKV(const MultiplePositions& batch_start,
                                const size_t num_interleaved) {
     PROFILER_ZONE("Gen.Attention.QKV");
     // For the computation of Q, K, and V, it is useful to remember that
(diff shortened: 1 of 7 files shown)


gemma/gemma-inl.h
```diff
@@ -848,32 +866,18 @@ void RangeChecks(size_t& max_tokens, size_t& max_generated_tokens,
...
 // Placeholder for internal test3, do not remove
 
-// Returns interleaved tokens: one from each query, followed by the second from
-// all queries, with EOS padding.
-static std::vector<int> InterleaveQueries(const MultiplePromptsTokens& queries,
-                                          const RuntimeConfig& runtime_config,
-                                          size_t& min_prompt_size,
-                                          size_t& max_prompt_size) {
+// Count the minimum/maximum size of prompts for interleave queries.
+static void InterleaveQueries(const MultiplePromptsTokens& queries,
```
jan-wassenberg, line 870: Please keep the std::vector return for now until we can remove the last reference to it.

jan-wassenberg (changes requested):
Very nice, I like that you also used a typedef. Looks good.

Good catch that prompt() is unused. We do have some internal code that uses it, though. Would you mind reverting that change? We can then redo the removal once it's no longer referenced.

gemma/gemma-inl.h
ufownl: OK, let me drop that commit. :)

ufownl:
I've dropped that commit. :)

jan-wassenberg (approved):
Thanks :)

jan-wassenberg:
hm, we seem to have change markers (<<<<<) inside the argument list of TransformerLayer in gemma-inl.h, not in the pull request but in the code that copybara imported. Maybe the force-push clobbered history? Can you rebase and double-check the diff is clean? Worst case, create a new pull request?

ufownl:
I saw the newest commit of the dev branch was just added some test cases about gemma2 and small fixes, there shouldn't be any conflict in gemma-inl.h? I have rebased my local branch to the newest dev branch, and there was no conflict. Is the latest code not in the dev branch?

jan-wassenberg:
I agree there isn't actually a conflict, I think the issue is just that the branch is out of date. A `git pull -r` followed by a push should fix it :)

ufownl:
I've updated this branch. See if the problem still exists?

jan-wassenberg:
Thanks, rerunning the import :)

jan-wassenberg:
Sorry, still seeing the conflict :( Would you mind re-creating the pull request?

ufownl:
OK, I close this PR first.

PR #422, opened 2024-10-11
Fix compilation error of the weights compression tool

ufownl:
GCC returns the errors about the missing "typename", and Clang also returns the warnings about them.


jan-wassenberg (approved):
Thanks :) C++20 has relaxed this typename requirement such that it wouldn't be required, but we can certainly add it back.

PR #426, opened 2024-10-14
Add overloads of `Image::ReadPPM` method

ufownl:
Make it able to load image data from a `hwy::Span`.

### DIFF
--- a/paligemma/BUILD
+++ b/paligemma/BUILD
@@ -11,7 +11,10 @@ cc_library(
     name = "image",
     srcs = ["image.cc"],
     hdrs = ["image.h"],
-    deps = ["@highway//:hwy"],
+    deps = [
+        "@highway//:hwy",
+        "//compression:io",
+    ],
 )
 
 cc_test(
--- a/paligemma/image.cc
+++ b/paligemma/image.cc
@@ -14,6 +14,7 @@
 // limitations under the License.
 
 #include "paligemma/image.h"
+#include "compression/io.h"
 
 #include <stdint.h>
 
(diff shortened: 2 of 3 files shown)


jan-wassenberg (dismissed):
I'm curious what your use case for streams is?

ufownl:
> I'm curious what your use case for streams is?

For example, receive image data from the network and cache it in RAM, then load and parse it from RAM directly. :)

jan-wassenberg:
Got it, thanks. I'm not very familiar with network stacks, but haven't heard of std::istream being used in that context. Should we maybe move towards an in-memory interface where we pass in a hwy::Span<uint8_t> for the image bytes?

ufownl:
I added a new overload to make it able to load image bytes from a`hwy::Span`. This is achieved with a custom stream class.

jan-wassenberg:
Thanks for adding the Span overload. Wow, I haven't seen a custom stream buffer class before.
It seems like a fair amount of complexity. Wouldn't it be easier if the Span function was the main implementation, which reads uint8_t directly from the span, and if other callers have an istream of some kind, we can read bytes from stream into a vector and then pass that to the Span overload?

For our use case, this would also include reading from file. Note that io.h has a ReadFileToString helper function we could also use for that.

ufownl:
The first solution I considered was the one you mentioned. However, P6 format files consist of a text header and binary data. If the Span overload was the main implementation, we must implement the text parser ourselves to parse the string and integers in the text header. It seems more complex than reusing `std::istream` in the standard library.

jan-wassenberg:
Ah, I understand. One alternative we could consider is reusing some of the existing code from [JPEG XL](https://sourcegraph.com/github.com/libjxl/libjxl/-/blob/lib/extras/dec/pnm.cc?L265-267), in particular SkipWhitespace and ParseUnsigned. What do you think?

ufownl:
Yeah, if we want to support more image formats in the future, using a parsing library like this is a better choice.

ufownl:
I removed the `std::istream` overload and refactored the span version. Now, it directly parses the PPM format on the span.

jan-wassenberg (dismissed):
Wonderful, thank you for making this change :D

ufownl:
Sorry, I'm unfamiliar with Bazel, and it has been fixed. :)

jan-wassenberg (approved):
Thanks for making the change!

PR #432, opened 2024-10-19
Fix compilation errors of "compress_weights" target

ufownl:
This PR fixes the error:

```
error: cannot pass object of non-trivial type 'DistortionStats' through variadic method; call will abort at runtime [-Wnon-pod-varargs]
```

### DIFF
--- a/compression/compress.h
+++ b/compression/compress.h
@@ -444,8 +444,10 @@ class CompressStats {
   char padding_[64];  // prevent false sharing
 };
 #else
+class DistortionStats;
+
 struct CompressStats {
-  void Notify(...) {}
+  void Notify(const DistortionStats&) {}
   void NotifyIn(int) {}
   void Assimilate(const CompressStats&) {}
   void PrintAll() {}


jonpsy:
Thank you, I faced similar error when I was compiling this on Mac. Should we rather have an overload of this function?

ufownl:
@jonpsy Currently, there should be no need to add an overload. This function only accepts a `DistortionStats` parameter, just like in the `COMPRESS_STATS` conditional block.

jonpsy:
What's weird is, this feature already exists, see L401. Weirdly enough, `#define COMPRESS_STATS` has been manually set to 0 in [e29cd56](https://github.com/google/gemma.cpp/commit/e29cd566cf3367671e8f59419a04e308796a7c57).

@austinvhuang can you comment on this?

jan-wassenberg (approved):
Thank you for fixing this. Clang seemed to accept it without complaint.

Yes, COMPRESS_STATS is 0 and we don't actually call this function. Our style guide discourages forward declarations but I figure it's fine here.

PR #435, opened 2024-10-24
Add a compilation option to disable topology

ufownl:
I ran the latest version of gemma.cpp on one of our devices, which has the topology `16x1x2, using 16x1x2`. The running speed was far slower than the previous versions (with a single thread pool). Therefore, I made this PR to add an option to disable topology detection and create a single package (with one cluster and one node) in `NestedPools` to run it with the same performance as before on these devices.

### DIFF
--- a/util/threading.h
+++ b/util/threading.h
@@ -28,6 +28,10 @@
 #include "hwy/contrib/thread_pool/thread_pool.h"
 #include "hwy/contrib/thread_pool/topology.h"
 
+#ifndef GEMMA_DISABLE_TOPOLOGY
+#define GEMMA_DISABLE_TOPOLOGY 0
+#endif  // !GEMMA_DISABLE_TOPOLOGY
+
 namespace gcpp {
 
 // A slice of a 1D integer range such as the indices of packages or clusters.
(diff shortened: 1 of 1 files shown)


util/threading.h
```diff
@@ -104,6 +108,7 @@ class BoundedTopology {
       enabled_lps.Set(lp);
     }
 
+#if !GEMMA_DISABLE_TOPOLOGY
     if (HWY_LIKELY(!topology_.packages.empty())) {
       InitFromTopology(enabled_lps, package_slice, cluster_slice);
     }
```
jan-wassenberg, line 114: I think we could just put an `#endif` here - without InitFromTopology, we will not have filled packages_, which is what NumPackages checks, so we can use the same fallback.

jan-wassenberg (commented):
Thanks, I understand the desire for this change. FYI we will soon be updating the matmul to actually use these clusters and restore or hopefully further improve performance vs before. But I see no harm in keeping this option.

ufownl:
Now, it uses that fallback. :)

jan-wassenberg (approved):
Nice, thanks for making the change :)

PR #473, opened 2025-01-17
Add the missing `migrate_weights` target for CMake

ufownl:
As title.


jan-wassenberg (approved):
Thanks for adding :)

jan-wassenberg:
We have an internal failure in the merge infra, looking into it.

PR #493, opened 2025-02-08
Fix the link error when building `compress_weights` with Clang on macOS

ufownl:
Implement NUQ version of `LayerWeightsPtrs<Weight>::Reshape` using template specialization instead of conditional branch.

### DIFF
--- a/gemma/weights.cc
+++ b/gemma/weights.cc
@@ -257,8 +257,8 @@ void ModelWeightsStorage::CreateForType(Type weight_type,
   }
 }
 
-template <class Weight>
-void LayerWeightsPtrs<Weight>::Reshape(MatStorage* storage) {
+template <>
+void LayerWeightsPtrs<NuqStream>::Reshape(MatStorage* storage) {
   if (attn_vec_einsum_w.data() == nullptr) return;
 
   const size_t model_dim = layer_config.model_dim;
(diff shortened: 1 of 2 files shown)


jan-wassenberg (commented):
Thanks for fixing! I think a simpler/smaller but sufficient patch would be to add if *constexpr* to the IsSame line, then we wouldn't have to expose the Reshape implementation in weights.h?

ufownl:
> Thanks for fixing! I think a simpler/smaller but sufficient patch would be to add if _constexpr_ to the IsSame line, then we wouldn't have to expose the Reshape implementation in weights.h?

Ideally, it should be like this. But if we put the Reshape implementation into weights.cc, when building compress_weights with Clang on macOS, it complained about the link error below:

```
Undefined symbols for architecture x86_64:
  "gcpp::LayerWeightsPtrs<float>::Reshape(gcpp::MatStorageT<hwy::uint128_t>*)", referenced from:
      void hwy::ThreadPool::Run<gcpp::ModelWeightsPtrs<float>::AllocAndCopyWithTranspose(hwy::ThreadPool&, std::__1::vector<gcpp::MatStorageT<hwy::uint128_t>, std::__1::allocator<gcpp::MatStorageT<hwy::uint128_t>>>&)::'lambda'(unsigned long long, unsigned long)>(unsigned long long, unsigned long long, gcpp::ModelWeightsPtrs<float>::AllocAndCopyWithTranspose(hwy::ThreadPool&, std::__1::vector<gcpp::MatStorageT<hwy::uint128_t>, std::__1::allocator<gcpp::MatStorageT<hwy::uint128_t>>>&)::'lambda'(unsigned long long, unsigned long) const&) in compress_weights.cc.o
      void hwy::PoolTasks::CallClosure<gcpp::ModelWeightsPtrs<float>::AllocAndCopyWithTranspose(hwy::ThreadPool&, std::__1::vector<gcpp::MatStorageT<hwy::uint128_t>, std::__1::allocator<gcpp::MatStorageT<hwy::uint128_t>>>&)::'lambda'(unsigned long long, unsigned long)>(void const*, unsigned long long, unsigned long) in compress_weights.cc.o
```

This is probably because of differences in template instantiation between GCC and Clang. And it seems that including compress-inl.h in weights.h is not a good idea, so I moved the implementation of NUQ version into a template specialization and put it in weights.cc.

jan-wassenberg (approved):
Yes indeed, it's important to avoid compress-inl in weights.h. Thanks for making the change :)

PR #529, opened 2025-03-26
Refactor `WrapAndTokenize` to work properly with Gemma3

ufownl:
This was discussed in #523.

### DIFF
--- a/gemma/tokenizer.cc
+++ b/gemma/tokenizer.cc
@@ -121,6 +121,11 @@ void GemmaChatTemplate::Init(const GemmaTokenizer& tokenizer) {
   HWY_ASSERT(tokenizer.Encode("<start_of_turn>model\n", &sot_model_));
   eot_.reserve(2);
   HWY_ASSERT(tokenizer.Encode("<end_of_turn>\n", &eot_));
+  HWY_ASSERT(tokenizer.Encode("\n", &pali_sep_));
+  vlm_soi_.reserve(2);
+  HWY_ASSERT(tokenizer.Encode("\n\n<start_of_image>", &vlm_soi_));
+  vlm_eoi_.reserve(2);
+  HWY_ASSERT(tokenizer.Encode("<end_of_image>\n\n", &vlm_eoi_));
 }
 
 std::vector<int> GemmaChatTemplate::Apply(size_t pos,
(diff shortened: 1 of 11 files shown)


gemma/tokenizer.cc
```diff
@@ -114,57 +114,96 @@ bool GemmaTokenizer::Decode(const std::vector<int>& ids,
   return impl_->Decode(ids, detokenized);
 }
 
+GemmaChatTemplate::GemmaChatTemplate(const GemmaTokenizer& tokenizer) {
+  Init(tokenizer);
```
jan-wassenberg, line 118: Let's inline this? Just looking at the header, it's not clear whether Init should be called by users.

gemma/tokenizer.cc
```diff
@@ -114,57 +114,96 @@ bool GemmaTokenizer::Decode(const std::vector<int>& ids,
...
+                                 const ModelInfo& info, size_t pos,
+                                 const std::string& prompt,
+                                 size_t image_batch_size) {
+  std::vector<int> text_part;
+  HWY_ASSERT(tokenizer.Encode(prompt, &text_part));
+  std::vector<int> tokens;
+  switch (info.wrapping) {
+    case PromptWrapping::PALIGEMMA: {
+      std::vector<int> sep;
+      HWY_ASSERT(tokenizer.Encode("\n", &sep));
```
jan-wassenberg, line 181: I'm curious why we do not precompute these (also soi/eoi below) like in GemmaChatTemplate?

jan-wassenberg (changes requested):
Thanks, nice to wrap this in a class.

ufownl:
> I'm curious why we do not precompute these (also soi/eoi below) like in GemmaChatTemplate?

These parts, ported from my project, have been integrated into `GemmaChatTemplate`. Two different vision prompt wrapping are implemented in `WrapPali` and `WrapVLM` for flexibility, and usually we can use `WrapAndTokenize` for convenience.

jan-wassenberg (approved):
Thanks for making the change. LGTM

ufownl:
Seems something is going wrong, I just rebased `dev` branch to sync that crash fixing.

ufownl:
@jan-wassenberg Something seemed to go wrong, and it was hanging. I only rebased the `dev` branch to sync that crash fix.

jan-wassenberg:
Sorry this got stuck. I check and it's similar to the issue we had last time where there's a merge conflict with internal code. I will manually patch this in..

PR #539, opened 2025-04-12
Add --prompt flag for non-interactive mode

prajwalc22:
This change adds a --prompt command-line option that allows users to provide prompts directly without entering interactive mode, which is useful for scripting and automation.


jan-wassenberg:
Thanks for adding, this is nice to have. We are only able to merge into the dev branch, would you mind rebasing/replacing your PR to there?

prajwalc22:
Thanks for the update! I attempted to rebase onto the dev branch, but I ran into some conflicts. I’ll work on resolving them and update the PR shortly.

jan-wassenberg:
Yes, sorry about the merge conflict. We recently cleaned up the interface a bit. More on the way related to weights handling, but it will take a few more days to land.
Sounds good!

prajwalc22:
I've successfully rebased the changes onto the dev branch and resolved the conflicts. The PR is now ready for review.  Would you be able to approve the workflow run when you have a chance? Thank you in advance!

README.md
```diff
@@ -1,583 +1,27 @@
-# gemma.cpp
+---
+library_name: gemma.cpp
```
jan-wassenberg, line 2: A model card can be useful, but let's not overwrite the previous README.md with it :) Please revert this file, we can add another file if you like.

gemma/gemma_args.h
```diff
@@ -28,13 +28,205 @@
...
+  } else if (HWY_IS_UBSAN) {
+    return "ubsan";
+  } else if (HWY_IS_DEBUG_BUILD) {
+    return "dbg";
+  } else {
+    return "opt";
+  }
+}
+template <typename Derived>
+struct ArgsBase {
```
jan-wassenberg, line 59: hm, this was/still is defined in util/args.h, is it necessary to change the base class for adding the flag?
If you want to improve the ArgsBase, we can overwrite the one in util/args.h instead?

gemma/gemma_args.h
```diff
@@ -28,13 +28,205 @@
...
+
+    visitor(
+        eot_line, "eot_line", std::string(""),
+        "End of turn line. "
+        "When you specify this, the prompt will be all lines "
+        "before the line where only the given string appears.\n    Default = "
+        "When a newline is encountered, that signals the end of the turn.",
+        2);
+
+    visitor(prompt, "prompt", std::string(""),
```
jan-wassenberg, line 191: The new prompt flag seems a better fit for InferenceArgs? BTW I'm curious where you got ThreadingArgs? This is newly written code from a pull request. I would take care of integrating it when that PR lands, no need to copy it now/before then :)

gemma/run.cc
```diff
@@ -238,6 +249,22 @@ void Run(ThreadingArgs& threading, LoaderArgs& loader,
   KVCache kv_cache =
       KVCache::Create(model.GetModelConfig(), inference.prefill_tbatch_size);
 
+  if (!threading.prompt.empty()) {
```
jan-wassenberg, line 252: I'd prefer not to duplicate this code, which increases maintenance burden. Instead we can re-use the existing ReplGemma, and pass to GetPrompt the inference args (rather than individual fields), and check inside GetPrompt if prompt is not empty, and if so return that immediately.
Then to exit, we could introduce a flag outside of the `while(true)` loop that checks whether prompt was not empty, and if so, exit the loop. This adds another condition next to the existing `prompt_string[1] == 'q'`. What do you think?

jan-wassenberg (changes requested):

gemma/run.cc
prajwalc22: Thank you for the review! I'll make the following changes:

1. README.md changes:
   - I'll revert the README.md changes back to original
   - Create a separate MODEL_CARD.md for the model documentation

2. ArgsBase implementation:
   - I'll remove the duplicate ArgsBase implementation from gemma_args.h since it's already in util/args.h

3. Prompt flag location:
   - I'll move the prompt flag from ThreadingArgs to InferenceArgs where it fits better logically
   - Will update verbosity level to 1 since it's a user-facing option

4. run.cc simplification:
   - I'll modify the code to reuse the existing ReplGemma implementation
   - Will implement the suggested approach of checking prompt in GetPrompt and adding a loop exit condition for non-interactive mode
   - Remove the separate code path for non-interactive mode

 I'll push these changes soon.

jan-wassenberg: Sounds good, thanks!

gemma/run.cc
```diff
@@ -149,21 +161,32 @@ void ReplGemma(const ThreadingArgs& threading, const InferenceArgs& inference,
...
-    if (prompt_string.empty()) {
+
+    if (inference.prompt.empty() && prompt_string.empty()) {
       std::cout << "Use '%q' to quit.\n";
       continue;
     }
+    std::vector<int> prompt;
+    size_t prompt_size = 0;
+    size_t prefix_end = 0;
+    if constexpr (kVerboseLogTokens) {
```
jan-wassenberg, line 185: Looks like this will always be empty. Let's move it back after prompt has been set?

gemma/run.cc
```diff
@@ -192,10 +211,13 @@ void ReplGemma(const ThreadingArgs& threading, const InferenceArgs& inference,
...
-      for (int i = 0; i < prompt_size; ++i) {
-        fprintf(stderr, "DDD TOKEN %3d: %6d\n", i, prompt[i]);
-      }
+    if (have_image) {
+      runtime_config.image_tokens = &image_tokens;
+      prompt_size = prompt.size();
+      // The end of the prefix for prefix-LM style attention in Paligemma.
+      prefix_end = prompt_size;
+      // We need to look at all the tokens for the prefix.
+      runtime_config.prefill_tbatch_size = prompt_size;
```
jan-wassenberg, line 220: Why this change? It is OK to have smaller batches than the prompt. For prompts > 512 or so, we want to break them up into multiple batches. I think we can leave prefill_tbatch_size unchanged, no?

jan-wassenberg (changes requested):
Thanks for making the changes. Almost there :)

prajwalc22:
@jan-wassenberg I've addressed all the feedback from the review:

1. Removed the change to prefill_tbatch_size in the image handling code - it's no longer being set to prompt_size
2. Fixed the variable placement issue by moving the prompt vector initialization to after the prompt_string is set
3. Simplified the non-interactive mode logic
4. Maintained the prompt flag in InferenceArgs as requested

Ready for another review, thanks for your feedback!

jan-wassenberg (approved):
Thanks for making the changes. Nice!

jan-wassenberg:
:) YW, thanks for implementing it!
CI has found an issue: the new include of threading.h in run.cc requires a dep be added to bazel BUILD, or preferably just removing the include, which I think is unnecessary.
FYI I will be out for Easter until Tue.

prajwalc22:
I'll remove the unnecessary threading.h include from run.cc  implementation. I'll make that change right away.

Enjoy your Easter break! :)
See you  when you're back on Tuesday.

jan-wassenberg (approved):
Thanks for updating :)

PR #566, opened 2025-05-13
Fix the wrapping field of the deduced model config

ufownl:
It looks like a typo. :)

But, there is another issue when generating with an image:

```
[ Reading prompt ] Abort at gemma-inl.h:1086: Assert prefix_end_this_query == 0 || max_tbatch_size >= prefill_this_query: 
Aborted (core dumped)
```


jan-wassenberg (approved):
Thanks for fixing. ACK on the image, will have a look tomorrow.

PR #573, opened 2025-05-20
Minor fixes for ViT

ufownl:
* Fix the incorrect batch size setting for `prefill_activations` in `PrefillVit` function.
* Fix incorrect `prefix_end` and `prefill_tbatch_size` settings for PaliGemma and Gemma3 in the CLI frontend.
* Fix the output of the CLI frontend.

After fixing these issues, the assertion failure will no longer be triggered, but ViT still does not work properly. I compared the image tokens generated by the main and dev branches, and the values ​​of the corresponding positions are unequal. There should be issues with the weight processing of the ViT part. I tried to look into it, but I have no idea yet.

### DIFF
--- a/gemma/run.cc
+++ b/gemma/run.cc
@@ -136,7 +136,7 @@ void ReplGemma(const ThreadingArgs& threading, const InferenceArgs& inference,
     ++tokens_generated_this_turn;
     if (in_prompt) {
       if (inference.verbosity >= 1) {
-        std::cerr << "." << std::flush;
+        std::cout << "." << std::flush;
       }
       return true;
     } else if (config.IsEOS(token)) {
@@ -188,7 +188,6 @@ void ReplGemma(const ThreadingArgs& threading, const InferenceArgs& inference,
                                     .use_spinning = threading.spin};
     inference.CopyTo(runtime_config);
     std::vector<int> prompt;
-    size_t prompt_size = 0;
     size_t prefix_end = 0;
     if (have_image) {
       prompt = WrapAndTokenize(gemma.Tokenizer(), gemma.ChatTemplate(),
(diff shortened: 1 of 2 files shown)


gemma/run.cc
```diff
@@ -188,20 +188,20 @@ void ReplGemma(const ThreadingArgs& threading, const InferenceArgs& inference,
...
-      // The end of the prefix for prefix-LM style attention in Paligemma.
-      // See Figure 2 of https://arxiv.org/abs/2407.07726.
-      prefix_end = prompt_size;
-
-      // REMOVED: Don't change prefill_tbatch_size for image handling
-      // runtime_config.prefill_tbatch_size = prompt_size;
+      if (config.wrapping == PromptWrapping::PALIGEMMA) {
+        // The end of the prefix for prefix-LM style attention in Paligemma.
+        // See Figure 2 of https://arxiv.org/abs/2407.07726.
+        prefix_end = prompt_size;
```
jan-wassenberg, line 201: The prefix_end change makes sense, thanks.
But the prefill_tbatch_size change can result in huge batch sizes and should not be necessary. Results ought to be the same regardless of batch size. Can we revert that?

What I can imagine is that we should still call activations.SetBatchSize. This sets the (active) number of rows in each tensor, which is initially the batch size passed to the Activations ctor.

jan-wassenberg: The intention was to split up the matmuls because at some point, larger batch sizes are slower, and for the matmul itself, it doesn't matter whether we chunk it or do all at once. I agree that Prefix-LM wants to attend to all image tokens. What we are missing for the chunking to work is the online softmax also used for flash attention, which is on the roadmap.
So you are right, it seems for the moment our attention has to have all K vectors for all tokens. Let's add a comment mentioning this and the plan to support online softmax, after which we can lift this requirement.

jan-wassenberg (approved):
Thanks for pointing that out and adding the comment :)

PR #574, opened 2025-05-22
Fix the ViT weights loading

ufownl:
The shape of `qkv_einsum_b` is 2D `(3 * heads, QKV dims)`, and its backend pointer will be passed into `MatMul` to do element-wise addition via `PackedScale1` method, so it must have no padding.

I checked the implementation of `MatPtr` and its derived classes and found that the assertions about packed and scale are available only in debug builds, making these bugs very hard to catch. Should we change them to the version available in the release builds?

### DIFF
--- a/gemma/weights.h
+++ b/gemma/weights.h
@@ -224,7 +224,7 @@ struct LayerWeightsPtrs {
       func(TENSOR_ARGS(vit.attn_out_w, kMustRead));
       func(TENSOR_ARGS(vit.attn_out_b, kMustRead));
       func(TENSOR_ARGS(vit.qkv_einsum_w, kMustRead));
-      func(TENSOR_ARGS(vit.qkv_einsum_b, kMustRead));
+      func(TENSOR_ARGS(vit.qkv_einsum_b, kMustRead | TensorArgs::kNoPad));
       // MlpBlock.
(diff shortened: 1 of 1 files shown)


jan-wassenberg (approved):
Thanks for catching this. I will add a comment explaining the issue when back on Monday.
It's indeed vexed with DASSERT vs ASSERT, we've had some discussions about this. Unfortunately Griffin has some PackedScale1 in inner loops and always asserting there could be expensive. We could mostly fix that by hoisting that outside loops, especially for Conv1D.
As an alternative, I anyway recommend running asan/msan when debugging. Those count as debug builds. Or would you like to try defining `__clang_analyzer__` locally to hopefully force the `HWY_IS_DEBUG_BUILD` to 1?

ufownl:
> As an alternative, I anyway recommend running asan/msan when debugging. Those count as debug builds. Or would you like to try defining __clang_analyzer__ locally to hopefully force the HWY_IS_DEBUG_BUILD to 1?

I thought your internal tests would run asan/msan, so I didn't think in that direction at first, until I tracked `nan` in the output of some layers.

jan-wassenberg:
Unfortunately only for the unit tests, which didn't cover this. Anything involving the actual weights is only currently only run manually. I'd like to change that but it's not yet clear how we will do that (embedding in binary vs loading from network attached storage).

PR #588, opened 2025-06-03
Fix the broken VitAttention

ufownl:
VitAttention is still MHA and does not use an external KV cache. So I think we should check whether it is used for ViT and then allocate KV space for it.


jan-wassenberg (approved):
Thanks for fixing, sorry about that. I am working on reviving paligemma_test to hopefully prevent recurrences.

PR #593, opened 2025-06-04
Fix compilation error on G++ 9.4

ufownl:
It can be compiled using Clang, but using a non-constant reference as an output argument doesn't seem sensible.


jan-wassenberg (approved):
Makes sense, thanks :)

PR #607, opened 2025-06-11
Fix the issue of duplicate `pos`

ufownl:
I think it should be increased there; otherwise, the last token of the prompt and the first token generated will have the same `pos`.


ufownl:
This issue has been fixed by #609.

PR #735, opened 2025-10-16
Update CMakePresets.json for Windows build issue with Cmake 4.x 

Hitesh-ed:
Issue:
CMake 4.x fails to configure the SentencePiece dependency because its CMakeLists.txt declares cmake_minimum_required(VERSION 3.1), which is no longer compatible (it has to be atleaset 3.5 when using latest Cmake 4.x for build).

Solution:
Adding the following cache variable in CMakePresets.json to enforce modern policies automatically:
```
"cacheVariables": {
  "CMAKE_POLICY_VERSION_MINIMUM": "3.5"
}

# Pretends old CMake minimums below 3.5 are OK (fixes SentencePiece 3.1 issue automatically)
```

This ensures all developers can run "cmake --preset windows" without hitting legacy compatibility or deprecation issues.

### DIFF
--- a/compression/int-inl.h
+++ b/compression/int-inl.h
@@ -159,7 +159,6 @@ class IntCodec {
     const hn::Rebind<int32_t, decltype(df)> di32;
     const hn::Rebind<int16_t, decltype(di32)> di16;
     const hn::Rebind<int8_t, decltype(di16)> di8;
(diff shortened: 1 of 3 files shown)


jan-wassenberg:
Nice, thanks for adding. We are only able to accept changes to the dev branch, would you please retarget the PR?

PR #827, opened 2026-01-14
Refactor: Move io to gemma/io and fix build paths

otesh-o:
I'm really sorry about the previous attempt and the noise i attracted. I didn't verify the build properly before pushing. I already did a full refactor to get it right.

I've moved the io directory into gemma/io as requested. To make sure this is actually solid, I’ve also updated the paths across both Bazel and CMake and renamed the binary to gemma_main to avoid the ArtifactPrefixConflictException
I’ve verified this builds 100% on Ubuntu, macOS, and Windows on my own fork to ensure I’m not breaking anything for others this time.
You can see the passing CI runs here: https://github.com/olaotesile/gemma.cpp/actions/runs/20975584510

Excited to learn more!


jan-wassenberg:
Apology accepted.
I appreciate your efforts here, and it looks like the build is indeed green :) But it's not clear to me why we should move io to gemma/io? The thinking with the current directory structure is that io is model-agnostic, and gemma would be the model-specific parts.
I didn't see any TODO or comment proposing this and I'm inclined to close the pull request because it's not clear what this move would improve.

treasure567:
baba wan be core contributor by force 😂😂

PR #833, opened 2026-02-12
Support building with Emscripten

brendandahl:
Update CMake configuration and utility functions to enable compilation
with Emscripten. This includes setting Wasm-specific flags like
memory64 and SIMD, implementing platform-specific memory detection, and
adding guards for features like OpenSSL that may be unavailable in a
web environment.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,20 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+if(EMSCRIPTEN)
+  add_compile_options("-sMEMORY64")
+  add_compile_options("-msimd128")
+  add_compile_options("-pthread")
+  add_link_options("-sALLOW_MEMORY_GROWTH")
+  add_link_options("-sMAXIMUM_MEMORY=16GB")
+  add_link_options("-sNODERAWFS")
+  add_link_options("-sMEMORY64")
+  add_link_options("-sSTACK_SIZE=2MB")
+  add_link_options("-pthread")
+  add_link_options("-sPROXY_TO_PTHREAD")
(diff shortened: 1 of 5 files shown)


brendandahl:
For a little more context, we'd like to start using gemma.cpp in some experimental Wasm AI benchmarks. If you're interested, I can also add CI testing for this configuration and help keep it green if needed.

jan-wassenberg (dismissed):
Nice. Benchmark sounds interesting, is there any info you can share?
Sure, we'd welcome a PR to add the CI and your help keeping it green.

brendandahl:
There's some more info on the benchmarks in [this repo](https://github.com/GoogleChrome/webai-compute-benchmark). Currently, it's mainly transformers.js (ONNX) and LiteRT workloads, but we'd like to add some more. We're still trying to figure out if we want to put Gemma.cpp in the default suite (what contributes to the overall score) since downloading such a big model would be slow. Gemma.cpp can at least go in the experimental benchmarks that devs can look at though.

We're planning to use the benchmarks to guide generating optimized Wasm SIMD from LLVM, improve v8 performance, and help see the impact of the FP16 proposal for Wasm.

In regards to the model size, is there a smaller model (<2B) that could be used for the benchmark that would simulate a similar workload?

jan-wassenberg:
Sounds interesting, thanks for sharing.
Yes, there is a 270M version on Kaggle that I'd recommend for reducing download size: https://www.kaggle.com/models/google/gemma-3/gemmaCpp/3.0-270m-it-sfp

FYI this PR is stuck on a permissions issue which I am fixing in cl/871333540.

jan-wassenberg:
We can have another go at merging this, the internal error is now resolved. Can you help rebase this PR?

brendandahl:
I've updated the branch, but ran into #885 yesterday.

jan-wassenberg:
Thanks :) Let's land this one first before investigating #885.

jan-wassenberg (approved):

PR #896, opened 2026-04-15
Fix integer overflow in tensor dimensions and PPM parser

sharadboni:
## Summary

- **Tensor dimension overflow (`util/basics.h`):** `Extents2D::Area()` computes `rows * cols` without checking for `size_t` overflow. When tensor dimensions are deserialized from a model file via `MatPtr::VisitFields()`, attacker-controlled `rows` and `cols` values can silently overflow, causing undersized allocations and subsequent heap buffer overflows. Added an overflow check that aborts if `cols > SIZE_MAX / rows`.

- **PPM image size overflow (`paligemma/image.cc`):** `ReadPPM()` computes `data_size = width * height * 3` without overflow protection. A crafted PPM file with large width/height values can overflow this computation, leading to an undersized `data_` vector and out-of-bounds writes during pixel parsing. Added a two-step overflow check before the multiplication.

- **PPM integer parsing overflow (`paligemma/image.cc`):** `ParseUnsigned()` accumulates decimal digits into a `size_t` without detecting overflow, allowing silently wrapped values to bypass subsequent size checks. Added overflow detection that returns `nullptr` (parse failure) when the accumulated value would exceed `SIZE_MAX`.

## Test plan
- [ ] Verify existing tests still pass (`bazel test //...`)
- [ ] Confirm PPM files with normal dimensions load correctly
- [ ] Confirm malformed PPM with overflow-inducing dimensions is rejected gracefully
- [ ] Confirm model files with overflow-inducing tensor dimensions trigger abort

### DIFF
--- a/paligemma/image.cc
+++ b/paligemma/image.cc
@@ -140,11 +140,11 @@ bool Image::ReadPPM(const hwy::Span<const char>& buf) {
   }
   ++pos;
   if (width == 0 || height == 0) {
-    std::cerr << "Invalid zero dimension\n";
+    HWY_ABORT("Invalid zero dimension\n");
     return false;
   }
   if (width > SIZE_MAX / 3 || width * 3 > SIZE_MAX / height) {
-    std::cerr << "Image dimensions overflow\n";
+    HWY_ABORT("Image dimensions overflow\n");
     return false;
   }
   const size_t data_size = width * height * 3;
--- a/paligemma/image.cc
+++ b/paligemma/image.cc
@@ -83,8 +83,11 @@ const char* ParseUnsigned(const char* pos, const char* end, size_t& num) {
   }
   num = 0;
   for (; pos < end && std::isdigit(*pos); ++pos) {
-    num *= 10;
-    num += *pos - '0';
+    const size_t digit = *pos - '0';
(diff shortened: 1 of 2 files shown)


sharadboni:
@jan-wassenberg Could you review this security fix? It adds overflow checks for tensor dimensions in Extents2D::Area() and fixes integer overflow in the PPM image parser that can be triggered via crafted model/image files.

paligemma/image.cc
```diff
@@ -137,6 +140,14 @@ bool Image::ReadPPM(const hwy::Span<const char>& buf) {
     return false;
   }
   ++pos;
+  if (width == 0 || height == 0) {
+    std::cerr << "Invalid zero dimension\n";
```
jan-wassenberg: Prefer HWY_ABORT here and below rather than std::cerr.

jan-wassenberg (changes requested):
Thanks, this seems reasonable to rule out. We're only able to land patches targeting the dev branch, would you mind updating the PR?
Also one code suggestion.

jan-wassenberg (approved):
Thanks for updating! (FYI HWY_ABORT does not return, but we can land as-is)

PR #919, opened 2026-05-21
Add direct HuggingFace safetensors loader for Gemma 4 (E2B/E4B)

ssfdre38:
## Summary

Adds a new code path to load Gemma 4 weights directly from HuggingFace `*.safetensors` files, bypassing the BlobStore conversion step. This avoids any potential weight precision loss from format conversion and lets you use freshly downloaded HF checkpoints without a separate conversion tool.

### New files
- `io/safetensors.h` / `io/safetensors.cc` — `SafetensorsIndex` class: scans a directory for `*.safetensors` shards, parses the 8-byte LE header + JSON, and provides `ReadTensor()` via seek-based I/O. Handles both single-file and sharded (`model.safetensors.index.json`) checkpoints.
- `gemma/load_safetensors.cc` — `WeightsPtrs::LoadFromSafetensors()`: maps HF tensor names to gemma.cpp `MatPtr` fields. Handles Q/K/V concat, gate/up-proj concat, o_proj direct load, and per-layer token embedding transpose (`[V, L*D]` → `[L*D, V]`). Calls `Fixup()` at the end.

### Modified files
- `gemma/weights.h` — adds `LoadFromSafetensors()` public declaration
- `gemma/model_store.h` / `model_store.cc` — adds `ModelStore(const ModelConfig&, const Path& tokenizer_path)` constructor for the safetensors path (reads tokenizer directly from file, leaves `scales_` empty)
- `gemma/gemma.h` / `gemma.cc` — adds `Gemma(ModelConfig, tokenizer, safetensors_dir, InferenceArgs, ThreadingContext)` constructor; changes `BlobReader reader_` to `unique_ptr<BlobReader>` to allow null when not using BlobStore
- `gemma/gemma_args.h` — adds `--safetensors` (directory path) and `--model_spec` (e.g. `gemma4-e4b-bf16-it`) flags to `LoaderArgs`
- `gemma/run.cc` — wires new flags into `Run()` with a conditional branch (uses `unique_ptr<Gemma>` to avoid copy/move)
- `CMakeLists.txt` — adds new source files; links `nlohmann_json` to `libgemma`

## Usage

```bash
./gemma --safetensors /path/to/gemma-4-e4b-it \
        --model_spec   gemma4-e4b-bf16-it \
        --tokenizer    /path/to/tokenizer.model \
        --prompt       "Hello!"
```

The `--model_spec` specifier uses the existing `ModelConfig(std::string)` format: `{model-prefix}-{type}-{wrapping}` e.g. `gemma4-e2b-bf16-it` or `gemma4-e4b-bf16-it`.

## Tested

- Builds cleanly (MSVC + ninja) with the existing CMake setup
- `--help` shows both new flags with descriptions
- E4B multimodal HF checkpoint (`model.language_model.*` prefix, 2130 tensors, 42 layers): loads fully, prompt processing begins
- Both single-shard and multi-shard (`*.index.json`) layouts supported by `SafetensorsIndex`


ssfdre38:
im part of cla already

jan-wassenberg:
Thanks for adding :)
I think the issue with the CLA is the co-author Copilot. We can work around that.

The bigger question is whether we should integrate other formats in this way. I'd prefer to have a single loader (from SBS), but we can imagine bundling converters that (offline) convert something like safetensors to SBS. We already have python/convert_from_safetensors.py - would you mind updating that to meet your needs?

Happy to land the Gemma4 changes - FYI we are already working on the MoE variant.

ssfdre38:
Thanks for the quick review!

On the CLA — happy to squash and remove the Copilot co-author trailer from the commits, that should sort it.

On the architecture question, that makes total sense — keeping a single loader path (SBS) is cleaner long-term. I was going the direct loader route mainly because I wanted to avoid a conversion step when iterating on new HF checkpoints, but an offline converter that produces SBS is just as good.

I can update \python/convert_from_safetensors.py\ to handle Gemma4 (E2B/E4B). The existing script is PaliGemma-only, so I'd add Gemma4 support — same Q/K/V concat + gate/up-proj concat logic I have in \load_safetensors.cc\, just emitting SBS instead.

A couple of quick questions before I rework it:
1. Should Gemma4 live in the same script (with a \--model\ flag to pick the variant), or a separate \convert_gemma4.py\?
2. Since you're already working on the MoE variant internally — is there a target tensor naming schema I should match to avoid conflicts when that lands?

Happy to pivot this PR to the converter approach once I have that context.

jan-wassenberg:
:) Sorry to reply late. Glad to hear you're fine with the offline converter step.
I'd trust your judgment on how much of the existing code can be reused from paligemma vs gemma4. I can imagine it's quite different, which would make me lean towards a separate file.
On tensor naming, we are using: `moe_gate_ein,moe_linear_w,post_ffw1_ns,post_ffw2_ns,router_scale`.
For non-MoE: `gating_ein,gating1_w,gating2_w`.

PR #924, opened 2026-05-29
[Security] Add missing blob-size assertion in ReadAllToBF16 weight loader

0xASTRA:
## Summary

`ReadAllToBF16` in `weights.cc` is the only one of three weight-load modes missing the `range.bytes == mat.PackedBytes()` assertion. The other two modes (`MapAll` at line ~557, `MakeBatches` at line ~645) both assert this invariant. Without the check, a crafted `.sbs` file with correct `Rows`/`Cols` metadata but an inflated blob `bytes` field passes all structural validation and causes a heap OOB write (`keep_type` branch) or OOB read (`DecompressToBF16` branch).

## Fix

Add `HWY_ASSERT_M(tensor.range.bytes == mat.PackedBytes(), mat.Name())` at the top of the per-tensor lambda in `ReadAllToBF16`, before either read path executes — mirroring the existing checks in `MapAll` and `MakeBatches`.

## Test plan

- [ ] Existing weight-loading tests pass
- [ ] A `.sbs` file with inflated blob `bytes` for one tensor is rejected with an assertion failure in `kReadBF16` mode (`--to_bf16` or `decode_qbatch_size >= 128`)


jan-wassenberg:
Thanks for making the fix :) We're only able to merge on the dev branch, would you mind retargeting to that?

0xASTRA:
> Thanks for making the fix :) We're only able to merge on the dev branch, would you mind retargeting to that?

Done!

jan-wassenberg (approved):

0xASTRA:
Thanks! Can it be closed?

jan-wassenberg:
FYI the logic here is a bit off. We only want to check this if keep_type == true, i.e. move the code down by one line. I'll fix that soon.

PR #947, opened 2026-07-08
Add initial T5Gemma S/S support

### LINKED ISSUE #661: T5Gemma support
No program supports T5Gemma model. Please could you add its support here or better in llama.cpp?

maasir554:
## Summary

Adds initial T5Gemma S/S support to gemma.cpp.

This includes:
- T5Gemma S/S model config and tensor metadata
- HF safetensors -> SBS conversion through `python/convert_from_safetensors.py`
- Encoder-decoder runtime path for fresh seq2seq generation
- BF16 default conversion path for correctness/parity
- Basic docs and focused config/tensor/cache tests

## Scope

This PR targets T5Gemma v1 S/S only.

Out of scope:
- T5Gemma 2
- Other T5Gemma sizes
- Multiturn/cache reuse with changing encoder inputs
- Fully optimized T5-specific attention/logits kernels

## Local Validation

- `bazel test -c opt //:configs_test //:tensor_info_test //:kv_cache_test`
- Local optimized smoke test with `t5gemma-s-s-it`
  - prompt: `Hello`
  - output begins: `I am a new user, ...`
  - local speed: roughly 30-35 generated tok/s on MacBook M1, 8GB RAM

## Notes
Opened as draft to get early maintainer feedback on scope, naming, and whether this should be split further. 

Additional Note: `//:gemma_lib_test` currently fails locally for me in an existing `GemmaAttentionFlash/EMU128` golden mismatch on Apple Silicon; this reproduced outside the T5Gemma changes.

Fixes #661 
Note: this initial PR adds T5Gemma S/S support. Additional T5Gemma variants can be added in follow-up PRs after this lands.

### DIFF
--- a/compression/python/compression_clif_aux.cc
+++ b/compression/python/compression_clif_aux.cc
@@ -52,12 +52,6 @@ HWY_BEFORE_NAMESPACE();
 namespace gcpp {
 namespace HWY_NAMESPACE {
 
-ThreadingArgs SingleThreadArgs() {
-  ThreadingArgs args;
-  args.max_lps = 1;
-  return args;
-}
-
 // Implementation for the currently compiled SIMD target.
 class SbsWriterImpl : public ISbsWriter {
   template <typename Packed>
@@ -98,7 +92,7 @@ class SbsWriterImpl : public ISbsWriter {
 
  public:
   SbsWriterImpl(const std::string& sbs_path)
-      : ctx_(SingleThreadArgs()), writer_(gcpp::Path(sbs_path), ctx_) {}
+      : ctx_(ThreadingArgs()), writer_(gcpp::Path(sbs_path), ctx_) {}
 
   void Insert(const char* name, F32Span weights, Type type,
               const TensorInfo& tensor_info) override {
--- a/compression/python/compression_clif_aux.cc
+++ b/compression/python/compression_clif_aux.cc
@@ -52,6 +52,12 @@ HWY_BEFORE_NAMESPACE();
 namespace gcpp {
 namespace HWY_NAMESPACE {
 
+ThreadingArgs SingleThreadArgs() {
+  ThreadingArgs args;
+  args.max_lps = 1;
+  return args;
+}
+
 // Implementation for the currently compiled SIMD target.
 class SbsWriterImpl : public ISbsWriter {
   template <typename Packed>
@@ -92,7 +98,7 @@ class SbsWriterImpl : public ISbsWriter {
 
  public:
   SbsWriterImpl(const std::string& sbs_path)
-      : ctx_(ThreadingArgs()), writer_(gcpp::Path(sbs_path), ctx_) {}
+      : ctx_(SingleThreadArgs()), writer_(gcpp::Path(sbs_path), ctx_) {}
 
   void Insert(const char* name, F32Span weights, Type type,
               const TensorInfo& tensor_info) override {
--- a/gemma/activations.h
+++ b/gemma/activations.h
@@ -367,7 +367,8 @@ struct Activations {
   Activations(const RuntimeConfig& runtime_config, const ModelConfig& config,
               size_t batch_size, size_t seq_len, ThreadingContext& ctx,
               std::vector<hwy::AlignedFreeUniquePtr<uint8_t*[]>>& row_ptrs)
-      : layer_config(config.layer_configs[0]),
+      : layer_config(config.is_encoder_decoder ? config.decoder_layer_configs[0]
+                                                : config.layer_configs[0]),
         moe_layer_config(MoELayerConfig(config)),
         mla_dims(config),
 
@@ -391,9 +392,15 @@ struct Activations {
         ple_token_emb(config.num_layers * config.ple_dim),
 
         max_workers(ctx.pools.MaxWorkers()),
-        s_ffw_in(config.num_layers, max_workers),
-        s_ffw_hidden(config.num_layers, max_workers),
-        s_ffw_out(config.num_layers, max_workers),
+        s_ffw_in(config.is_encoder_decoder ? config.decoder_num_layers
+                                            : config.num_layers,
+                 max_workers),
+        s_ffw_hidden(config.is_encoder_decoder ? config.decoder_num_layers
+                                               : config.num_layers,
+                     max_workers),
+        s_ffw_out(config.is_encoder_decoder ? config.decoder_num_layers
+                                             : config.num_layers,
+                  max_workers),
         router_in(MatFactory("router_in",
                              MoEBatchSize(moe_layer_config, batch_size),
                              config.model_dim, ctx.allocator)),
@@ -448,17 +455,41 @@ struct Activations {
                                               : config.rope_theta,
             config.yarn_factor, config.yarn_orig_seq_len, config.yarn_beta_fast,
             config.yarn_beta_slow)),
-        s_router_in(config.num_layers, max_workers),
-        s_router_logits(config.num_layers, max_workers),
-        s_expert_in(config.num_layers, max_workers),
-        s_expert_hidden(config.num_layers, max_workers),
-        s_expert_out(config.num_layers, max_workers),
-        s_w_expert_in1(config.num_layers, max_workers),
-        s_w_expert_in2(config.num_layers, max_workers),
-        s_w_expert_hidden(config.num_layers, max_workers),
-        s_w_gating_einsum_w1(config.num_layers, max_workers),
-        s_w_gating_einsum_w2(config.num_layers, max_workers),
-        s_w_linear_w(config.num_layers, max_workers),
+        s_router_in(config.is_encoder_decoder ? config.decoder_num_layers
+                                              : config.num_layers,
+                    max_workers),
+        s_router_logits(config.is_encoder_decoder ? config.decoder_num_layers
+                                                  : config.num_layers,
+                        max_workers),
+        s_expert_in(config.is_encoder_decoder ? config.decoder_num_layers
+                                              : config.num_layers,
+                    max_workers),
+        s_expert_hidden(config.is_encoder_decoder ? config.decoder_num_layers
+                                                  : config.num_layers,
+                        max_workers),
+        s_expert_out(config.is_encoder_decoder ? config.decoder_num_layers
+                                               : config.num_layers,
+                     max_workers),
+        s_w_expert_in1(config.is_encoder_decoder ? config.decoder_num_layers
+                                                 : config.num_layers,
+                       max_workers),
+        s_w_expert_in2(config.is_encoder_decoder ? config.decoder_num_layers
+                                                 : config.num_layers,
+                       max_workers),
+        s_w_expert_hidden(config.is_encoder_decoder ? config.decoder_num_layers
+                                                    : config.num_layers,
+                          max_workers),
+        s_w_gating_einsum_w1(config.is_encoder_decoder
+                                  ? config.decoder_num_layers
+                                  : config.num_layers,
+                              max_workers),
+        s_w_gating_einsum_w2(config.is_encoder_decoder
+                                  ? config.decoder_num_layers
+                                  : config.num_layers,
+                              max_workers),
+        s_w_linear_w(config.is_encoder_decoder ? config.decoder_num_layers
+                                               : config.num_layers,
+                     max_workers),
         attention_impl(runtime_config.attention_impl),
         attention_storage(config, layer_config, batch_size, seq_len,
                           runtime_config, ctx.pools.MaxWorkers(), ctx.allocator,
--- a/gemma/activations.h
+++ b/gemma/activations.h
@@ -373,6 +373,7 @@ struct Activations {
         mla_dims(config),
         num_layers(config.is_encoder_decoder ? config.decoder_num_layers
                                              : config.num_layers),
+        encoder_seq_len(config.is_encoder_decoder ? seq_len : 0),
 
         x(MatFactory("x", batch_size, config.model_dim, ctx.allocator)),
         x_bf(MatFactory("x_bf", batch_size, config.model_dim, ctx.allocator)),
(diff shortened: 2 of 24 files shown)


gemma/activations.h
```diff
@@ -368,26 +369,56 @@ struct Activations {
             MatFactory("ffw_out", batch_size, config.model_dim, ctx.allocator)),
 
         max_workers(ctx.pools.MaxWorkers()),
-        s_ffw_in(config.num_layers, max_workers),
-        s_ffw_hidden(config.num_layers, max_workers),
-        s_ffw_out(config.num_layers, max_workers),
+        s_ffw_in(config.is_encoder_decoder ? config.decoder_num_layers
```
jan-wassenberg: Let's add a member variable initialized to `config.is_encoder_decoder ? config.decoder_num_layers : config.num_layers`, then we can use that everywhere.

gemma/configs.h
```diff
@@ -438,6 +440,14 @@ struct ModelConfig : public IFields {
     visitor(use_global_timescale);
     visitor(partial_rotary_factor);
 
+    visitor(is_encoder_decoder);
+    visitor(encoder_num_layers);
```
jan-wassenberg: We can make `encoder/decoder*` fields conditional, i.e. only send them if is_encoder_decoder == true. (It's enough to wrap them in an if().)

gemma/gemma.cc
```diff
@@ -145,6 +147,836 @@ static float EmbeddingScaling(size_t model_dim) {
...
+    HWY_ASSERT(weights_t->Cols() == model_dim);
+    const auto embedding_span =
+        MakeSpan(weights_t->Row(0), embedding_ofs + model_dim);
+    DecompressAndZeroPad(hn::ScalableTag<float>(), embedding_span,
+                         embedding_ofs, x.Row(x_row), model_dim);
+    MulByConst(emb_scaling * weights_t->Scale(), x.Row(x_row), model_dim);
+  });
+}
+
+static constexpr int kT5GemmaPadId = 0;
```
jan-wassenberg: Let's move these to tokenizer.h? We can also reuse BOS_ID = 2.

gemma/gemma.cc
```diff
@@ -145,6 +147,836 @@ static float EmbeddingScaling(size_t model_dim) {
...
+                                               size_t key_pos) {
+  if (cache.pad_mask[key_pos]) return false;
+  const size_t window_size = T5GemmaEncoderWindowSize(config, layer_idx);
+  if (window_size >= config.max_seq_len) return true;
+  const size_t distance = query_pos > key_pos ? query_pos - key_pos
+                                              : key_pos - query_pos;
+  return distance <= window_size;
+}
+
+static HWY_INLINE float T5GemmaMaybeSoftCap(float score, float cap) {
```
jan-wassenberg: Reuse MaybeLogitsSoftCap from ops-inl.h?

gemma/gemma.cc
```diff
@@ -145,6 +147,836 @@ static float EmbeddingScaling(size_t model_dim) {
...
+    EmbedTokenFromWeights(prompt[pos], pos, config,
+                          weights.t5gemma_encoder_embedding,
+                          encoder_cache.hidden_states);
+    HWY_DASSERT(encoder_cache.pad_mask[pos] ==
+                (prompt[pos] == kT5GemmaPadId ? 1 : 0));
+  }
+
+  const LayerConfig& layer_config = config.encoder_layer_configs[0];
+  const size_t source_len = encoder_cache.source_len;
+  MatStorageT<BF16> pre_att_rms_out(
```
jan-wassenberg: Allocating is pretty expensive, probably we want to hoist this into something like Activations?

gemma/gemma.cc
```diff
@@ -145,6 +147,836 @@ static float EmbeddingScaling(size_t model_dim) {
...
+  const size_t heads_per_kv = heads / kv_heads;
+
+  RMSNormBatched(activations.x, layer.pre_cross_attention_norm_scale,
+                 activations.attention.pre_att_rms_out, env.ctx);
+  CallMatMul(activations.attention.pre_att_rms_out, layer.cross_q_einsum_w,
+             /*add=*/nullptr, env, activations.attention.q);
+
+  for (size_t qi = 0; qi < qbatch.Size(); ++qi) {
+    const T5GemmaEncoderCache* encoder_cache = qbatch.T5EncoderCache(qi);
+    HWY_ASSERT(encoder_cache != nullptr);
```
jan-wassenberg: Note that our HWY_ASSERT is always on, HWY_DASSERT only in debug/sanitizer builds. Might want to switch to HWY_DASSERT here?

jan-wassenberg (changes requested):
Remarkably nice work, very clean code, congrats :) A few suggestions:

maasir554:
> Remarkably nice work, very clean code, congrats :) A few suggestions:

Thank you, really appreciate the review and kind words :)
 I’ll address your suggestions in a follow-up update.

gemma/activations.h
maasir554: sure, I'll use a `size_t` variable  named `num_layers` for this.

gemma/configs.h
maasir554: Okay, in my update, only `visitor(is_encoder_decoder)` will be serialised, and rest all specifics in the `if` block.

maasir554:
Hi, I addressed the suggestions in the latest push:

- added a shared `Activations::num_layers`
- made encoder/decoder config fields conditional on `is_encoder_decoder`
- moved/reused T5 token IDs via `tokenizer.h` / `BOS_ID`
- reused `MaybeLogitsSoftCap`
- hoisted T5Gemma encoder scratch buffers into `Activations`
- switched internal encoder-cache checks to `HWY_DASSERT`

Thanks again for the review!

compression/python/compression_clif_aux.cc
```diff
@@ -51,6 +51,12 @@ HWY_BEFORE_NAMESPACE();
 namespace gcpp {
 namespace HWY_NAMESPACE {
 
+ThreadingArgs SingleThreadArgs() {
```
jan-wassenberg: What's the rationale for disabling threading here? It can help for very large files.

gemma/activations.h
```diff
@@ -366,28 +369,76 @@ struct Activations {
                       ctx.allocator)),
         ffw_out(
             MatFactory("ffw_out", batch_size, config.model_dim, ctx.allocator)),
+        t5_encoder_pre_att_rms_out(MatFactory(
+            "t5_e_pre_att", config.is_encoder_decoder ? seq_len : 0,
```
jan-wassenberg: In a similar way to num_layers, you might want to add an encoder_seq_len which is `config.is_encoder_decoder ? seq_len : 0`, then we can use that everywhere.
The other uses of `config.is_encoder_decoder ?` are less frequent, so maybe can leave as-is.

gemma/configs.cc
```diff
@@ -753,6 +807,14 @@ bool ModelConfig::OverwriteWithCanonical() {
 
 Model DeduceModel(const Path& blob_path, size_t layers, int layer_types) {
   switch (layers) {
+    case 8:
+      if (layer_types & kDeducedT5Gemma) {
+        return Model::T5GEMMA_S_S;
+      }
+      HWY_WARN("Failed to deduce model type from %s, layer count %zu types %x.",
```
jan-wassenberg: Rather than duplicate the warning logic here, I think it would be better to `break;` with a comment like // unknown, then move the warning logic out of the default: statement, and place it after the end of the switch }.

gemma/weights.cc
```diff
@@ -165,6 +165,92 @@ void LayerWeightsPtrs::SplitAttW1() {
...
+  const size_t qkv_dim = layer_config.qkv_dim;
+
+  att_weights.SetType(attn_vec_einsum_w.GetType());
+  HWY_ASSERT(att_weights.Rows() == model_dim);
+  HWY_ASSERT(att_weights.Cols() == heads * qkv_dim);
+  HWY_ASSERT(attn_vec_einsum_w.Rows() == heads * model_dim);
+  HWY_ASSERT(attn_vec_einsum_w.Cols() == qkv_dim);
+
+  {
+    static std::mutex m;
```
jan-wassenberg: The code on which this was based has been fixed, please also use the same pattern here:
https://github.com/google/gemma.cpp/commit/6241d5232aa90f784f5cd2ae6b40e5280a0b8e87#diff-513210d58714f91b767de71eccf6fcfac9df6f935ca098eb1316c2c09d080217
(no extra static mutex here, but reuse g_mat_owners_mutex.

python/BUILD.bazel
```diff
@@ -34,7 +34,6 @@ pybind_extension(
 py_binary(
     name = "run_example",
     srcs = ["run_example.py"],
-    strict_deps = False,
```
jan-wassenberg, line 37: What was the issue here? Our tooling may reintroduce such flags again.

jan-wassenberg (commented):
Thanks for updating :) A few more questions and one fix:

compression/python/compression_clif_aux.cc
maasir554: Good point, I don’t have a strong reason to force single-threading here.

I originally added it while debugging local conversion behavior, but it is not necessary for the final path and would indeed hurt large-file conversion. I removed `SingleThreadArgs()` and switched the writer back to the default `ThreadingArgs()`.

While validating this, I also hit an unrelated import failure from a duplicated `PostNormType.Scale` Python binding, so I fixed that too. `//compression/python:compression_test` now passes locally.

python/BUILD.bazel
maasir554: No strong issue here. 
I removed it while cleaning up the Python Bazel target during local converter work, but it is unrelated to T5Gemma support. 
Since this may be intentionally managed by tooling, I restored `strict_deps = False`.

maasir554:
Thanks a lot for the detailed review and suggestions :)

I’ve pushed a follow-up commit addressing the comments. A couple of these were cleanup mistakes on my side, especially the temporary single-threading change and the unrelated Bazel flag removal. I’m still new to this codebase, so the review was really helpful for understanding the project conventions better.

Thanks again for taking the time to explain the preferred patterns.

jan-wassenberg:
Thanks for making the changes. LGTM. We've just landed a big change to add Deepseek v4 which touched several files. Would you mind rebasing?

maasir554:
Thanks! I noticed the DeepSeek V4 changes yesterday :)

I’ve rebased my branch on the latest upstream/dev and pushed the updated PR. Please let me know if anything else needs adjustment.

gemma/configs.h
pculliton: Did you mean to lose the if() block here?

gemma/weights.cc
pculliton: Seems like this might have reverted inadvertently. Might be good to check your code for anything else that might have been dropped.

pculliton (changes requested):
Hi! Thanks for the PR. Jumping in to help Jan with final reviews here. Looks like some earlier fixes might have been lost. Would you mind doing a pass to make sure you're not losing older changes you'd like to keep?

maasir554:
Hi @pculliton ! Thanks for checking! I re-checked the current branch after the rebase.

The encoder/decoder config fields are still guarded by `if (is_encoder_decoder)`, and the weights.cc mutex change is using the shared `g_mat_owners_mutex` now. 

I think ypur comments may be attached to an **outdated diff** after the rebase/force-push, but please let me know if you’re seeing a remaining case I missed.

Screenshot 2026-07-22 at 8 38 19 PM
Screenshot 2026-07-22 at 8 40 12 PM

gemma/configs.h
maasir554: @pculliton as mentioned in the later conversation, the `if()` block is still there in the latest push.
Let me know you are still seeing the older version, might be some github / backend specific issue.
I'll retry to force push in that case.

pculliton (dismissed):
> I think ypur comments may be attached to an **outdated diff** after the rebase/force-push, but please let me know if you’re seeing a remaining case I missed.

Yes! Thanks - sorry, was looking at an outdated diff. Appreciate the double-check. Can't see any other cases.

I think eventually it will make sense to move T5- (and other model-) specific code out of gemma.cc, but for now as long as your testing is working, this all LGTM.

maasir554:
No worries, thanks for double-checking!

> I think eventually it will make sense to move T5- (and other model-) specific code out of gemma.cc, but for now as long as your testing is working, this all LGTM.

Agreed. I also think that would be the better long-term direction as more model families are added. While working on this, I was wondering whether some of the model-specific behavior/configuration could eventually move behind a cleaner model-family abstraction, instead of accumulating in gemma.cc.

I’ll keep this PR focused on the current T5Gemma S/S support, but I’d be happy to help with that cleanup separately if maintainers think it makes sense.

maasir554:
Also, I rebased again on current latest upstream/dev after the native tokenizer change and pushed the updated branch. Local `configs_test`, `kv_cache_test`, and `//:gemma` build pass.

jan-wassenberg (approved):

jan-wassenberg:
Import in progress :)

maasir554:
Thank you! This would be my first merged PR here, so I'm really excited. 
I appreciate all the review guidance along the way :)

jan-wassenberg:
:)
FYI several internal patches were added and this PR will show up as merged soon, when those are done.

maasir554:
Thank you so much! This was an amazing learning experience for me, and I really appreciate all the guidance and review feedback throughout the PR.

I’d be happy to come back with follow-up PRs for more T5Gemma & T5Gemma-2 variants if that sounds useful.

Thanks again :)

jan-wassenberg:
Thanks for your warm words. Am impressed by your work :)
I think additional variants would be less urgent, it is good to have the capability for testing/experimentation, but we don't have to support all variants.

PR #959, opened 2026-07-19
Add DeepseekV4 Flash support: inference, converter, MTP speculative decoding

iStrzalka:
Adds DeepseekV4 Flash support: inference, converter, and MTP speculative decoding.

Tested via running the model to check whether output made sense.


gemma/configs.cc
```diff
@@ -495,6 +495,7 @@ static ModelConfig ConfigGemma4_26B_MoE() {
   return config;
 }
 
+<<<<<<< ours
```
jan-wassenberg: Merge conflict?

gemma/configs.h
```diff
@@ -223,7 +259,11 @@ enum class Model {
   GEMMA3_12B_LM,
   GEMMA3_27B_LM,
   GEMMA4_26B_MOE,
+<<<<<<< ours
```
jan-wassenberg: Here also.

gemma/configs.h
```diff
@@ -560,9 +748,36 @@ struct ModelConfig : public IFields {
 
   InternalModelConfig internal;
   bool use_global_timescale = false;  // for Gemma 3
+<<<<<<< ours
```
jan-wassenberg: Several in this file.

gemma/gemma.cc
```diff
@@ -406,18 +415,26 @@ static HWY_NOINLINE void PrefillTBatch(const ModelConfig& config,
 
       // Fill activations.x (much faster than TransformerLayer).
       size_t image_token_position = 0;
+<<<<<<< ours
```
jan-wassenberg: Another :)

gemma/kv_cache.cc
```diff
@@ -80,6 +80,7 @@ KVCache::KVCache(const Extents2D& kv_extents, size_t num_layers,
   // fires. The 2-arg constructor path relies on k_v_cols == 0 to skip reshape.
 }
 
+<<<<<<< ours
```
jan-wassenberg: Another :)

jan-wassenberg (changes requested):
Interesting, thanks for adding! Excellent work.
Looks like we have some merge conflicts in kv_cache.cc, gemma.cc, configs.h/cc, weights.h, python/configs.cc - please resolve them to unblock merge. They seem independent, we can keep both.

jan-wassenberg (approved):
Thanks for updating!

PR #963, opened 2026-07-23
Add build flag to enable profiler

RaviTriv:
PR to float idea of an easy way to enable profiler via build flags and document how to use.

`bench_matmul` with `--config=profiler`
image


jan-wassenberg (approved):
Nice idea, thanks :)

PR #967, opened 2026-07-27
Fix and document the `--attention_impl` flag

RaviTriv:
Ran into some issues when messing around with using `--attention_impl` flag, proposed changes to document and fix :)

- Allocate `k_tile_vec`, `v_tile_vec` and init them in `AttentionActivationsPtrs` to fix segfault.

`./build/gemma --weights ~/development/gemma-weights/1b/1b-it-sfp.sbs --attention_impl flash_transposed_qs_bf16 --prompt "Where is Brampton?"`

**Current** 
image

**Updated**
image

- Pass flags through to eval harnesses


jan-wassenberg (approved):
Good changes, thanks!

PR #970, opened 2026-07-29
Integrate oneDNN matmul threadpool backend

romirdes:
## Summary

Adds an optional oneDNN matmul-primitive backend for `MatMul`, built against oneDNN's **THREADPOOL** CPU runtime so oneDNN parallelizes by calling back into gemma.cpp's existing thread pool instead of spawning its own threads.

The backend is opt-in and off by default:

- Bazel: `bazel build --define gemma_onednn_matmul=1 ...`
- CMake: `cmake -DGEMMA_ONEDNN_MATMUL=ON ...`

It is mutually exclusive with the existing BRGeMM backend (`GEMMA_ONEDNN_BRGEMM`), because oneDNN's CPU runtime (SEQ vs THREADPOOL) is a whole-library compile-time choice. Enabling both fails fast: a CMake `FATAL_ERROR`, an ambiguous `select()` in Bazel, and an `#error` guard in `ops/onednn_matmul.h`.

The path engages only for BF16×BF16 with `M > 1` (M == 1 measured slower than stock gemma for some shapes) and falls through to the standard MatMul whenever oneDNN declines or fails, so behavior is unchanged when the flag is off.

## Changes

| File | Description |
| --- | --- |
| `ops/onednn_matmul.h` | New public header: build-flag guard, mutual-exclusion `#error`, and the threadpool adapter that lets oneDNN dispatch work onto gemma.cpp's thread pool. |
| `ops/onednn_matmul-inl.h` | New `DoMatMul_OneDnn()` implementation: builds/caches the matmul primitive and reordered weights, applies scale and `add`, returns false to decline unsupported cases. |
| `ops/matmul-inl.h` | Dispatch into the oneDNN path for BF16×BF16 with `M > 1`, guarded by `GEMMA_ONEDNN_MATMUL`, with fall-through to the existing MatMul. |
| `ops/matmul.h` | Export `ops/onednn_matmul.h`; add `MMPerKey::onednn_built` to record that the oneDNN path handled a shape. |
| `ops/bench_matmul.cc` | Exclude the first oneDNN call (JIT + weight reorder) from timings, since its cost is amortized by the primitive/weights caches; oneDNN has no autotune sweep. |
| `util/zones.h` | Add `Callers::kOneDnnMatMul` profiler zone. |
| `util/zones.cc` | Map the new zone to the `"OneDnnMatMul"` name. |
| `BUILD.bazel` | Add the `gemma_onednn_matmul` config setting, `GEMMA_ONEDNN_MATMUL=1` define, new headers, and the `@onednn_tp` dependency. |
| `MODULE.bazel` | Add the `@onednn_tp` archive — same oneDNN v3.11 tarball/sha256 as `@onednn` (shared download cache), different `build_file` for the THREADPOOL runtime. |
| `bazel/onednn.bzl` | New shared build logic: `onednn_targets(cpu_runtime)` derives the runtime-dependent config lines and conditional copt from its argument. |
| `bazel/onednn.BUILD` | De-duplicated down to a single `onednn_targets(cpu_runtime = "SEQ")` call. |
| `bazel/onednn_threadpool.BUILD` | New build file: a single `onednn_targets(cpu_runtime = "THREADPOOL")` call. |
| `CMakeLists.txt` | Add the `GEMMA_ONEDNN_MATMUL` option, the mutual-exclusion `FATAL_ERROR`, and a `FetchContent` oneDNN v3.11 build configured with `DNNL_CPU_RUNTIME=THREADPOOL`. |

## Testing

Built and benchmarked with `ops/bench_matmul` under both Bazel and CMake, against the stock MatMul and the BRGeMM backend. Default (flag-off) builds are unaffected.


ops/onednn_matmul-inl.h
```diff
@@ -0,0 +1,205 @@
...
+#include "hwy/highway.h"
+
+HWY_BEFORE_NAMESPACE();
+namespace gcpp {
+namespace HWY_NAMESPACE {
+
+#if GEMMA_ONEDNN_MATMUL
+
+// Thread-local byte buffer for oneDNN's user-managed scratchpad.
+inline std::vector<uint8_t>& GetOneDnnScratchpad() {
```
jan-wassenberg: We could have multiple independent instances within one binary. Let's move this into MatMulEnv.per_cluster in matmul.h; it can be indexed by cluster_idx. (We can have concurrent calls to MatMul, one per cluster, PER MatMulEnv.)

ops/onednn_matmul-inl.h
```diff
@@ -0,0 +1,205 @@
...
+  static thread_local std::vector<uint8_t> scratch;
+  return scratch;
+}
+
+// oneDNN data type for the C element type (only f32/bf16 outputs are used).
+template <typename TC>
+constexpr dnnl::memory::data_type OneDnnDstType() {
+  if constexpr (hwy::IsSame<TC, float>()) {
```
jan-wassenberg: Nit: like IsBF16, let's use IsF32 here - that will correctly handle const TC.

ops/onednn_matmul.h
```diff
@@ -0,0 +1,154 @@
...
+    return h;
+  }
+};
+
+struct OneDnnWeightsEntry {
+  dnnl::memory packed;
+};
+
+inline auto& GetOneDnnWeightsCache() {
+  static std::unordered_map<OneDnnWeightsKey, OneDnnWeightsEntry,
```
jan-wassenberg: This should also reside in MatMulEnv to allow independent instances.

jan-wassenberg (changes requested):
Very nice work :) Let's move the state into MatMulEnv as mentioned below, otherwise LGTM.

ops/onednn_matmul-inl.h
romirdes: [cb32c29](https://github.com/google/gemma.cpp/pull/970/commits/cb32c29824098d18974592fa14d2a19fae6feb5b)

ops/onednn_matmul.h
romirdes: [d476c4b](https://github.com/google/gemma.cpp/pull/970/commits/d476c4bc3f6854f0aee744a01e130d7f61f1fc71)

ops/onednn_matmul-inl.h
romirdes: [bfce5c2](https://github.com/google/gemma.cpp/pull/970/commits/bfce5c27235959f7619d13edcc3d90d8accfb400)

jan-wassenberg (approved):
Apologies, I am running behind on reviews :/

jan-wassenberg (approved):

PR #980, opened 2026-08-16
Report dependency download progress during CMake configure

### LINKED ISSUE #979: Bad build UX: CMake configure is silent for minutes while cloning dependencies
A cold `cmake --preset make` prints nothing between sentencepiece's status lines and the end of configure:

```
-- VERSION: 0.2.1
-- Not Found TCMalloc: TCMALLOC_LIB-NOTFOUND
        <-- many minutes of silence here
```

FetchContent sets `FETCHCONTENT_QUIET=ON` by default, so the `git clone` of the remaining dependencies produces no output at all. There is no way to tell whether the build is working or has stalled, and no indication of how much is left.

It is not a short wait. On a slow link, `nlohmann/json` (~105k objects) took **639s of a 669s configure**, entirely without output.

Ideally the configure would name the dependency it is fetching and let git's own transfer progress through.

Mikyx-1:
Fixes #979.

## Problem

A cold configure goes silent for minutes while FetchContent clones dependencies, with no way to tell whether it is working or stalled. Measured on a slow link, `nlohmann/json` took 639s of a 669s configure without printing anything.

## Change

Adds `cmake/GemmaFetch.cmake`. Its `gemma_fetch()` wraps `FetchContent_Declare` + `FetchContent_MakeAvailable` with a progress bar naming each dependency and its elapsed time, and turns `FETCHCONTENT_QUIET` off so git's own `Receiving objects` counter provides the fine-grained progress CMake cannot know about.

```
-- [========------------] 40% | dependency 3/5: json -- fetching and configuring, please wait
Receiving objects:  59% (62581/104950), 116.54 MiB | 30.00 KiB/s   <-- updates live
--          json ready (639s)
```

`-DGEMMA_QUIET_FETCH=ON` restores the previous silent behavior for CI logs.

Used by the top-level build and by both examples, which are standalone projects with their own fetches.

## Notes

- The examples fall back to a plain no-progress `gemma_fetch()` if the folder is copied out of the tree, so they still configure standalone.
- The module is deliberately not `include_guard()`ed: an example fetching gemma.cpp as a subproject gets its own counter reset, so the inner build counts `1/5..5/5` instead of continuing the outer numbering.
- Clone semantics are unchanged. `GIT_SHALLOW` was considered and rejected: CMake implements it as `git clone --depth 1 --no-single-branch` + `git checkout <sha>`, which fails for the SHA-pinned dependencies. Fetching source tarballs instead would cut the json wait to seconds, but that is a separate change.

## Testing

- Cold configure of the top-level build: all 5 dependencies fetched, exit 0.
- `compile_commands.json` verified to still carry the sentencepiece and httplib include paths (the wrapper is a function, so populated `*_SOURCE_DIR` variables are hoisted to `PARENT_SCOPE`); `tokenizer/sentencepiece_tokenizer.cc` and `gemma/api_server.cc` compile clean.
- `examples/hello_world` configured with gemma.cpp as a subproject, confirming the nested counter reset.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,8 +60,15 @@ if(EMSCRIPTEN)
   add_link_options("-sEXIT_RUNTIME=1")
 endif()
 
-FetchContent_Declare(highway GIT_REPOSITORY https://github.com/google/highway.git GIT_TAG 9d5b12611fcfe145f988771c45e7bae9f78cb7fa EXCLUDE_FROM_ALL)
-FetchContent_MakeAvailable(highway)
+# Dependency fetching. gemma_fetch() reports download progress so the configure
+# step does not sit silent for minutes; see cmake/GemmaFetch.cmake.
+set(GEMMA_DEP_TOTAL 5)
+if(GEMMA_ONEDNN_BRGEMM OR GEMMA_ONEDNN_MATMUL)
+  math(EXPR GEMMA_DEP_TOTAL "${GEMMA_DEP_TOTAL} + 1")
+endif()
+include(${CMAKE_CURRENT_LIST_DIR}/cmake/GemmaFetch.cmake)
+
+gemma_fetch(highway GIT_REPOSITORY https://github.com/google/highway.git GIT_TAG 9d5b12611fcfe145f988771c45e7bae9f78cb7fa EXCLUDE_FROM_ALL)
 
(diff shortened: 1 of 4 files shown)


jan-wassenberg (approved):
Neat, thanks for adding :)

PR #981, opened 2026-08-16
fix temperature flag so it takes effect

RaviTriv:
Currently the temperature flag is not taking effect since its applied to the exponentiated logits and cancels against the normalization. Updating to divide the logits by temperature before exp so it affects the output.

## Current
**Temperature 0**
image

**Temperature 2**
image

## Updated
**Temperature 0**
image

**Temperature 2**
image

image

### DIFF
--- a/ops/ops-inl.h
+++ b/ops/ops-inl.h
@@ -1350,7 +1350,7 @@ static HWY_NOINLINE void Softmax(Logits logits, ThreadingContext& ctx,
                                  const SMOptions& sm_options = {}) {
   GCPP_ZONE(ctx, worker, Zones::kOpsSoftmax);
   HWY_DASSERT(logits.size() != 0);
-  HWY_ASSERT(temperature >= 0.0f);
+  HWY_DASSERT(temperature >= 0.0f);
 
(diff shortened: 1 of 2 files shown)


ops/ops-inl.h
```diff
@@ -1352,6 +1350,7 @@ static HWY_NOINLINE void Softmax(Logits logits, ThreadingContext& ctx,
                                  const SMOptions& sm_options = {}) {
   GCPP_ZONE(ctx, worker, Zones::kOpsSoftmax);
   HWY_DASSERT(logits.size() != 0);
+  HWY_ASSERT(temperature >= 0.0f);
```
jan-wassenberg: This runs quite often, so prefer HWY_DASSERT.

jan-wassenberg (changes requested):
Thanks for fixing. We generally run with deterministic mode because our sampler is very basic, but good to fix this. One nit:

ops/ops-inl.h
RaviTriv: good catch

jan-wassenberg (approved):

Issue #560, opened 2025-05-09
Optimization: Convert Hidden States to Bfloat16


FabianSchuetze:
In line with #543  and #164 I have worked a bit with changing the precision of the `pre_att_rms_out` from float to BF16. First, as also suggested, I made only a tiny code change. The patch would be:
```
diff --git a/gemma/activations.h b/gemma/activations.h
index 86345e2..a994e98 100644
--- a/gemma/activations.h
+++ b/gemma/activations.h
@@ -41,7 +41,7 @@ struct Activations {
   RowVectorBatch<float> logits;
 
   // Attention
-  RowVectorBatch<float> pre_att_rms_out;
+  RowVectorBatch<BF16> pre_att_rms_out;
   RowVectorBatch<float> att;      // attention vector
   RowVectorBatch<float> att_out;  // attention output
   // Accumulation of attention outputs over heads
@@ -88,7 +88,7 @@ struct Activations {
       logits = RowVectorBatch<float>(Extents2D(batch_size, vocab_size));
     }
 
-    pre_att_rms_out = RowVectorBatch<float>(Extents2D(batch_size, model_dim));
+    pre_att_rms_out = RowVectorBatch<BF16>(Extents2D(batch_size, model_dim));
     att = RowVectorBatch<float>(
         Extents2D(batch_size, heads * weights_config.seq_len));
     att_out = RowVectorBatch<float>(Extents2D(batch_size, heads * qkv_dim));
diff --git a/gemma/gemma-inl.h b/gemma/gemma-inl.h
index ccb34f0..e7eece4 100644
--- a/gemma/gemma-inl.h
+++ b/gemma/gemma-inl.h
@@ -293,7 +293,7 @@ class GemmaAttention {
         // Proceed row by row because there will be wraparound.
         for (size_t interleaved_idx = 0; interleaved_idx < num_interleaved;
              ++interleaved_idx) {
-          const float* x = activations_.pre_att_rms_out.Batch(interleaved_idx);
+          const hwy::bfloat16_t* x = activations_.pre_att_rms_out.Batch(interleaved_idx);
           const size_t query_idx = interleaved_idx % num_queries_;
           const size_t batch_idx = interleaved_idx / num_queries_;
           KVCache& kv_cache = kv_caches_[query_idx];
```
I wonder if this change was too small? 

[...]

Issue #164, opened 2024-04-26
Near-term roadmap


jan-wassenberg:
We're sharing a roadmap of ideas for improving and speeding up Gemma. If you'd like to join in and help us get there faster, please reach out so we can coordinate :)

### Threading
1) (jan-wassenberg) detect: CPU name, total #logical, per-logical: package, chiplet, core, smt, L2D/L3 size
2) CCX-aware pinning
3) (jan-wassenberg) more efficient ThreadPool

### Dot product
* Add `_mm*_dpbf16_ps` to `HWY_AVX3_SPR` and `HWY_AVX3_ZEN4` targets, plus define `HWY_NATIVE_DOT_BF16` in `set_macros-inl.h`
* Add new `NEON_*` target that uses `vbfdot` for `ReorderWidenMulAccumulate`
* If `!defined(HWY_NATIVE_DOT_BF16) || !HWY_NATIVE_DOT_BF16`, decompress bf16->f32 to temp array before MatVec

### Matmul
1) (pculliton) implement basic matmul and test. Not using BLAS because we want to fuse matmul and decompression.
2) Update Prefill to use matmul (activation @ weights) instead of MatVec
3) Cache-aware tiling

### Compression
1) (pculliton, ...) Eval infrastructure
2) decide whether NUQ is launchable
3) improved compression, possibly error diffusion

### General infra
* (pculliton) Python wrapper
* (pculliton, ...) Improved CI: run on Kaggle infra

MathiasSchindler:
Is Paligemma part of the scope of gemma.cpp?

jan-wassenberg:
Let's discuss in #185 :)

Gopi-Uppari:
Could you please confirm if this issue is resolved for you with the above comment ? Please feel free to close the issue if it is resolved ?

Thank you.

PR #987, opened 2026-08-22
Store pre_att_rms_out as BF16

Mikyx-1:
`pre_att_rms_out` is only ever the A operand of the QKV MatMuls, and there is no f32 MatMul kernel — `MaybeDecompressA` converts it to BF16 on every call, with no caching across calls. Storing it as BF16 lets `RMSNormBatched` write the final type directly and removes that pass.

Its siblings `pre_ffw_rms_out`, `x_bf` and `att_sums` are already BF16, so this applies the *"change most activations to bf16"* item from the #164 roadmap to one more buffer.

This is the same change proposed in #560 by Fabian Schuetze, rebased onto the current activations. That patch no longer applies (it predates `MatStorageT`, and had to hand-fix a `const float* x`); today every consumer — `RMSNormBatched`, `LayerNormBatched`, `CallMatMul` — is already generic over the element type, so it reduces to two type changes.

## Honest framing: this is cleanup, not a speedup

The redundant work is measurably gone. Profiler build, Apple M2, gemma2-2b-sfp-pt, 926-token prefill + 64 generated:

| | `MM.DecompressA` calls | time |
|---|---|---|
| before | 50,930 | 11.9 ms |
| after | 14,816 | 3.5 ms |

The binary also shrinks 656 KB, since the f32-A MatMul instantiations for these call sites are no longer emitted, and the buffer's footprint halves (37.7 → 18.9 MB at `prefill_tbatch 4096`).

But end-to-end there is **no measurable speedup**, matching the null result reported in #560. Interleaved A/B, 4 reps each: prefill 32.96 → 33.28 tok/s (σ 1.20), total 34.80 → 34.08 s. That's inside one standard deviation.

It can't be otherwise: the conversion is O(M·K) sitting in front of an O(M·K·N) matmul, so it is inherently ~1/N of the work it precedes (N ≈ 2560 here). Worth noting the null result is uniform across architectures — #560 saw it on both a non-bf16 x86 laptop and a Cortex-X3 *with* native bf16, and this is an M2 with `NEON_BF16`. It isn't a bf16-hardware split.

## No new hardware bf16 requirement

`MaybeDecompressA` returns `StridedViewBF` in both branches, so the kernel receives the same type and runs identical instructions either way. Targets without `HWY_NATIVE_DOT_BF16` widen to f32 via `PromoteEvenTo`/`PromoteOddTo` exactly as they did before — that widening is a property of the kernel and the CPU, not of how this buffer is stored. What goes away is one pass over M×K, on every target.

## Behaviour is unchanged

- 96 generated tokens are **byte-identical** before and after (temperature 0, 926-token prompt, gemma2-2b-pt).
- `gemma_test` cross entropy identical to every printed digit: `1.108557`.
- `attention_test` goldens pass **unchanged** on `NEON_BF16`, `NEON_WITHOUT_AES` and `EMU128` — no golden regeneration needed.

`FillRandom` in `attention_test.cc` is templated on the element type so it can still fill the buffer; without this the Bazel build breaks.

## Testing limitations

- arm64 only; I have no x86 machine to hand. The change is type-only, but the `attention_test` tolerance is loose (`KV_t = BF16` → 3e-2), so that pass is not a tight numerical proof — the byte-identical generation is the stronger evidence.
- Only the Gemma2 text path was exercised at runtime. `pre_att_rms_out` also feeds DeepSeek MLA, T5Gemma and ViT/Gemma4-ViT; those compile but were not run.
- `tiled_attention_test` could not be built to verify — it needs gmock+absl and doesn't build under CMake at all (pre-existing). It uses the already-templated `FillMatPtrT`, so it should be unaffected.

Happy to drop this if it isn't worth the review cycle — the value is genuinely marginal and I'd rather not misrepresent it as a perf win.

### DIFF
--- a/gemma/activations.h
+++ b/gemma/activations.h
@@ -197,7 +197,9 @@ struct AttentionActivations {
   MatStorageT<KV_t> vit_K_T;
   MatStorageT<KV_t> vit_V_T;
 
-  MatStorageT<float> pre_att_rms_out;
+  // BF16 because this is only ever the A operand of the QKV MatMuls, which
+  // would otherwise decompress it to BF16 on every call; see `pre_ffw_rms_out`.
+  MatStorageT<BF16> pre_att_rms_out;
   MatStorageT<float> att_out;      // attention output
   MatStorageT<float> att_out_reps;  // attention output for each thread.
   MatStorageT<float> softmax_max;  // see OnlineSoftmaxState
@@ -308,7 +310,7 @@ struct AttentionActivationsPtrs {
   MatPtrT<KV_t> vit_V_T;
 
   // Output of RMSNorm before attention, size batch_size x model_dim.
-  MatPtrT<float> pre_att_rms_out;
+  MatPtrT<BF16> pre_att_rms_out;
   // Attention output computed from att * V, size batch_size x (q_heads *
   // qkv_dim).
   MatPtrT<float> att_out;
--- a/gemma/attention_test.cc
+++ b/gemma/attention_test.cc
@@ -43,12 +43,13 @@ HWY_BEFORE_NAMESPACE();
 namespace gcpp {
 namespace HWY_NAMESPACE {
 
-void FillRandom(MatPtrT<float>& mat, uint64_t seed) {
+template <typename T>
+void FillRandom(MatPtrT<T>& mat, uint64_t seed) {
   hwy::RandomState rng(seed);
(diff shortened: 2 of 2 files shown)


jan-wassenberg (approved):
Thanks! Binary size and memory savings are worthwhile, hence happy to land this.

PR #993, opened 2026-08-27
Compress the embedding table to SFP at load time

Mikyx-1:
## Summary

- add `--sfp_embedding` to convert the input embedding from F32/BF16 to SFP while loading
- use two-pass, 4 MiB chunked conversion to bound temporary memory and preserve the tensor scale
- disable automatic memory mapping when conversion is requested; explicit `--map 1` still takes precedence and warns
- validate the source blob size before chunked reads
- document that logits MatMul speedups apply to tied input/output embeddings, while untied models retain the memory saving
- add focused tests for F32/BF16 conversion, scale propagation, partial chunks, invalid blob sizes, and mode selection

This implements the embedding-compression approach discussed in #164. Existing SFP read and MatMul paths require no changes after loading.

## Results

Measured on an M2 with four workers using `2.0-2b-pt-sfp.sbs` in read mode:

| Metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| `Gen.EmbeddingMatmul` | 654 us/token | 468 us/token | -28.5% |
| Decode | 11.21 tok/s | 12.42 tok/s | +10.8% |
| Maximum RSS | 3482 MiB | 2928 MiB | -554 MiB |
| Cross entropy/byte | 1.372889 | 1.368894 | no degradation |

## Testing

- `cmake --build build --target weights_test -j4`
- `ctest --test-dir build -R '^WeightsTest\.' --output-on-failure`
- `cmake --build build --target gemma -j4`
- CLI smoke test with Gemma 3 270M, both automatic mapping and explicit `--map 1`

### DIFF
--- a/gemma/weights.cc
+++ b/gemma/weights.cc
@@ -29,18 +29,21 @@
 #include "gemma/configs.h"
 #include "gemma/gemma_args.h"
 #include "gemma/model_store.h"
+#include "gemma/weights_internal.h"
+#include "hwy/base.h"
+#include "hwy/highway.h"
+#include "hwy/profiler.h"
 #include "io/blob_store.h"
 #include "util/mat.h"
 #include "util/threading_context.h"
 #include "util/zones.h"
-#include "hwy/base.h"
-#include "hwy/highway.h"
-#include "hwy/profiler.h"
 
 // TODO: move into foreach_target
 #include "compression/compress-inl.h"
 
 namespace gcpp {
+
+using weights_internal::TensorToRead;
 
 static std::mutex g_mat_owners_mutex;
 
@@ -657,12 +660,19 @@ std::vector<uint32_t> WeightsPtrs::AddTensorDataToWriter(
 }
 
 // Decides whether to read or map based on heuristics and user override.
-static WeightsPtrs::Mode ChooseMode(uint64_t file_bytes,
-                                    const LoaderArgs& loader,
-                                    const InferenceArgs& inference,
-                                    const Allocator& allocator) {
+WeightsPtrs::Mode weights_internal::ChooseMode(uint64_t file_bytes,
+                                               const LoaderArgs& loader,
+                                               const InferenceArgs& inference,
+                                               const Allocator& allocator) {
   Tristate to_bf16 = loader.to_bf16;
   Tristate map = loader.map;
+
+  // An explicit request to convert the embedding requires owned memory. Do
+  // not let the automatic mapping heuristic override that request. An
+  // explicit --map=1 is still honored and diagnosed in ReadFromBlobs.
+  if (loader.sfp_embedding == Tristate::kTrue && map == Tristate::kDefault) {
+    map = Tristate::kFalse;
+  }
 
   // Disable mapping if not padded to the base page size.
   if (file_bytes % allocator.BasePageBytes() != 0) {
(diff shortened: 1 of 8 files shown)


gemma/weights.cc
```diff
@@ -836,6 +842,172 @@ static void ReadAllToBF16(const std::vector<TensorToRead>& tensors,
...
+              Callers::kReadAllToSFP, [&](uint64_t chunk, size_t thread) {
+                GCPP_ZONE(ctx, thread, Zones::kStartupWeightsReadAllToSFP);
+                const size_t begin = chunk * rows_per_chunk;
+                const size_t end = HWY_MIN(begin + rows_per_chunk, rows);
+                Chunk<T> buffers(cols, rows_per_chunk);
+                const float* raw =
+                    buffers.Decompress(tensor, reader, begin, end);
+
+                float maxabs = 0.0f;
+                for (size_t i = 0; i < (end - begin) * cols; ++i) {
```
jan-wassenberg: Is this worth vectorizing? It's nice to minimize startup time. Recommend two separate vectors and then ReduceMax(d, Max(v1, v2)).

gemma/weights_test.cc
```diff
@@ -0,0 +1,231 @@
...
+  return ThreadingContext(args);
+}
+
+template <typename T>
+static float SourceValue(const T value) {
+  return static_cast<float>(value);
+}
+
+template <>
+float SourceValue(const BF16 value) {
```
jan-wassenberg: Both conversions can just be hwy::ConvertScalarTo<float>.

jan-wassenberg (changes requested):
Nice work, thanks! Two minor suggestions.

gemma/weights.cc
Mikyx-1: Done in a075d4e. MaxAbs now uses two independent Highway vector accumulators and combines them with ReduceMax(df, Max(max0, max1)); the remaining vector and scalar tail are handled explicitly. On a Gemma 3 4B-sized BF16 chunk (819 x 2560 floats), the loop benchmark improved from 2.110 ms to 0.270 ms (~7.8x), with a bit-identical maximum. Thanks!

gemma/weights_test.cc
Mikyx-1: Done in 4b49a8f. Both SourceValue and MakeSourceValue now use hwy::ConvertScalarTo, so the explicit BF16 specializations are gone. Thanks!

gemma/weights.cc
```diff
@@ -836,6 +842,195 @@ static void ReadAllToBF16(const std::vector<TensorToRead>& tensors,
...
+  ParallelFor(
+      Parallelism::kFlat, num_chunks, ctx, /*cluster_idx=*/0,
+      Callers::kReadAllToSFP, [&](uint64_t chunk, size_t thread) {
+        GCPP_ZONE(ctx, thread, Zones::kStartupWeightsReadAllToSFP);
+        const size_t begin = chunk * rows_per_chunk;
+        const size_t end = HWY_MIN(begin + rows_per_chunk, rows);
+        Chunk<T> buffers(cols, rows_per_chunk);
+        const float* raw = buffers.Decompress(tensor, reader, begin, end);
+
+        using DF = hwy::HWY_NAMESPACE::ScalableTag<float>;
```
jan-wassenberg: Let's add `namespace hn = hwy::HWY_NAMESPACE;` (typically at namespace scope), then write hn::Zero to shorten the code :)

jan-wassenberg (changes requested):
Nice, congrats on the speedup. Interesting the compiler was not able to autovectorize this effectively.
Note that weights.cc does not yet use dynamic dispatch, but that's probably fine, there's only light compute happening here.

One more minor nit for readability:

gemma/weights.cc
Mikyx-1: Done in f3407af. Added namespace hn = hwy::HWY_NAMESPACE; at namespace scope and used the hn:: shorthand for the Highway operations, including hn::Zero. Also ran clang-format and verified all four weights_test cases pass. Thanks for giving me feedback!

jan-wassenberg (approved):
LGTM, thanks for updating!

PR #1012, opened 2026-09-04
skip kv cache allocation for kv-shared layers

RaviTriv:
Gemma4 has cross layer attention :). PR to reduce some memory allocation by skipping layers which will not need KV Cache allocated. Reducing allocation by 20 layers for Gemma4-2B ,  about 57% of KV cache


jan-wassenberg (approved):
Nice one, makes sense.

PR #1017, opened 2026-09-06
Reduce local-attention KV storage with per-layer rings

### LINKED ISSUE #1016: Reduce KV memory for local-attention layers
## Problem

On `dev` (`1658f88`), the default `flash` attention path reserves full-context BF16 KV/K/V storage for local and global layers alike. Local layers only attend to their sliding window, so retaining capacity for the entire context wastes memory as the configured context grows.

The runtime-aware cache constructor also allocates compact tiled buffers that the default Flash path does not use. The tiled attention backends already have compact local rings; the default Flash path still reads the legacy transposed K/V buffers.

## Fix direction

- Allocate separate BF16 buffers per owning layer for the default Flash path.
- Size local rings for the attention window plus the full prefill batch and trailing SIMD padding, capped by the configured context. The extra rows prevent batch writes and padding from overwriting history needed by early queries.
- Keep global layers at the configured context capacity, with the logical context limit separate from physical alignment padding.
- Reuse the source layer's buffers for shared-KV layers and retain the largest window required by their consumers.
- Preserve live history when runtime batches require larger rings, keep snapshots independent, and clear buffers safely when reusing a cache.
- Allocate only the buffers used by the selected attention backend and preserve the existing BF16 attention arithmetic and model-specific cache layouts.

Mikyx-1:
Closes #1016

Local-attention layers currently allocate KV storage for the entire context. This change gives each layer its own FP32 buffer: local layers use `min(seq_len, window - 1 + max(1, prefill_batch_size))` rows, and global layers keep full-context storage.

The extra batch rows prevent QKV prefill writes from overwriting history needed by earlier queries. Runtime growth preserves cached history, copies remain independent, and attention reads use each layer's ring capacity while score-buffer indexing and arithmetic remain unchanged.

Validation on Gemma 3 270M and 1B: all 10 A/B comparisons produced byte-identical decoded output and identical token IDs (1,280 generated tokens per build). Runs used a shared baseline-calibrated matrix plan to isolate the cache change from timing-dependent autotuning.

| Model | KV capacity, MiB | Peak process RAM, MiB | Prefill, tokens/s | Decode, tokens/s |
|---|---:|---:|---:|---:|
| 270M | 288.5 → 72.7 | 644.9 → 565.6 | 1013.3 → 1082.3 | 40.48 → 42.30 |
| 1B | 416.5 → 100.0 | 1574.3 → 1458.5 | 199.2 → 231.7 | 13.79 → 15.00 |

Medians from three alternating pairs with a 3,367-token prompt, 128 generated tokens, 8,192-token capacity, 256-token prefill batches, and six pinned workers on an i5-12400F. CPU load varied; speed improvements are measured observations, not a universal performance guarantee. RAM savings were consistent.

- CMake cache/attention regression tests: 8/8 passed, covering wraparound, runtime growth, prefix attention, global layers, context limits, copies, and exact FP32 attention comparisons on AVX2 and EMU128.
- AddressSanitizer with leak detection: 8/8 passed. Changed KV/attention translation units and tests were instrumented; supporting libraries were reused.
- CLI, single benchmark, API server, and batch benchmark built successfully.

C++ API change: direct users of the former public `KVCache::kv_cache` matrix must use `LayerCache(layer)` or `Row(layer, position)`. Repository callers are updated.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -130,6 +130,18 @@ cc_library(
 )
 
 cc_test(
+    name = "kv_cache_test",
+    srcs = ["gemma/kv_cache_test.cc"],
+    deps = [
+        ":configs",
+        ":gemma_args",
+        ":kv_cache",
+        ":threading_context",
+        "@googletest//:gtest_main",
+    ],
+)
+
+cc_test(
     name = "flash_attention_test",
     srcs = ["gemma/flash_attention_test.cc"],
     deps = [
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,6 +222,7 @@ set(GEMMA_TEST_FILES
   compression/sfp_test.cc
   evals/gemma_test.cc
   gemma/flash_attention_test.cc
+  gemma/kv_cache_test.cc
   gemma/tensor_info_test.cc
   io/blob_store_test.cc
   io/fields_test.cc
--- a/gemma/attention.cc
+++ b/gemma/attention.cc
@@ -50,14 +50,16 @@ namespace gcpp {
 namespace HWY_NAMESPACE {
 
 // Computes Q.K scores, which are "logits" (or scores) stored to att.
-// `k` is a strided view of the kv cache with dimensions [seq_len, qkv_dim].
+// `k` is a strided view of the kv cache with dimensions [cache_rows, qkv_dim].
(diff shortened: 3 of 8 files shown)


Mikyx-1:
Superseded by #1020, which ports the local-cache changes to dev and includes new dev-baseline validation.

PR #1019, opened 2026-09-06
Block FP32 vision attention value accumulation

### LINKED ISSUE #1018: main: reduce repeated value loads in Gemma 3 vision attention
The Gemma 3 vision attention path on `main` repeatedly loads V and updates the output vector for each source patch. This is a proposal to reduce that work within each image encoding.

**Scope**

Applies to `DotSoftmaxWeightedSumMatrix` in [`gemma/vit.cc` on main at 3ed403e](https://github.com/google/gemma.cpp/blob/3ed403e28707c0e3eb5f48b0e487b63a446d8e2e/gemma/vit.cc#L119-L131). Current `dev` uses a different FlashAttention implementation; this proposal targets the existing `main` path.

**Problem**

- The thread pool already parallelizes query rows, but each row independently traverses all source patches.
- Every source position calls `MulByConstAndAdd`, reloading V from the interleaved QKV layout and reading/writing the running output vector.
- Neighboring query rows use the same V, but the loop does not explicitly share those loads or retain output accumulators across source positions.

**Proposed solution**

- Pack each head's V into contiguous FP32 scratch storage alongside the existing K copy, reusing the buffer across heads.
- Compute four query rows together with a Highway SIMD kernel, sharing V loads and retaining two vectors of running sums per query in registers until the tile is complete.
- Preserve FP32 arithmetic and the original source-position accumulation order; handle remaining rows and partial SIMD vectors.
- Schedule groups of four rows through the existing thread pool and remove the redundant output-zeroing pass.

This adds one per-head V scratch matrix. It reuses data during a single encoding and requires no cache of previous images or prompts.

Mikyx-1:
Gemma 3 vision attention on `main` calls `MulByConstAndAdd` for every query/source pair, repeatedly loading V and updating the output vector. This change packs V per head and computes four query rows together, keeping FP32 accumulators in SIMD registers and sharing V loads.

Fixes https://github.com/google/gemma.cpp/issues/1018

**Scope and implementation**

- Targets `dev`. The current patch and measurements were developed against `main` at `3ed403e`; `dev` has replaced this loop with FlashAttention, so integration requires a port and fresh validation. The measurements below do not establish a speedup over `dev`.
- Preserves the source-position accumulation order, supports row/vector tails, and uses the existing thread pool.
- Adds 1.125 MiB of scratch for 4,096 patches × 72 FP32 values, reused across heads.
- Includes only the kernel, its integration, and required CMake/Bazel header entries.

**Measurement setup**

- CPU: Intel Core i5-12400F; six pinned physical CPU workers, AVX2, spin waits enabled.
- Platform/build: Linux x86-64, GCC 13.3.0, Release (`-O3 -DNDEBUG`).
- Model: Gemma 3 4B, `4b-it-sfp-padded.sbs` weights.
- Input: repository `paligemma/testdata/image.ppm`, resized to 896 × 896 (4,096 patches).
- Three serial A/B pairs, alternating order: before/after, after/before, before/after.
- Each process performed one warmup encoding and one measured encoding, with a common baseline-calibrated MatMul plan to isolate this change from autotuning differences.
- Both local A/B builds included the same pre-existing KV-cache changes; the baseline was not a pristine `main` build. Those changes are excluded from this PR.

**Image-encoder latency**

| Pair | Before | After |
|---|---:|---:|
| 1 | 38.019 s | 19.455 s |
| 2 | 38.899 s | 19.985 s |
| 3 | 38.887 s | 19.703 s |
| Median | **38.887 s** | **19.703 s** |

**1.97× faster image encoding; 49.3% lower latency (19.18 s saved per image).**
Timing excludes model/image loading, resizing, language-model prefill, text generation, and output-file writes. Each pass recomputes the image embeddings; there is no previous-image cache.

**Validation**

- Complete image embeddings were byte-identical in all three pairs: 256 × 2,560 FP32 values per run. Generated text was not compared.
- Local kernel comparisons passed 182 cases per target on AVX2 and EMU128, covering tails, unaligned/padded data, adjacent output heads, and varied values.
- All 12 related regression tests passed; all four kernel tests also passed AddressSanitizer/UndefinedBehaviorSanitizer with leak detection (kernel/tests instrumented).
- CLI and API server builds passed. The isolated PR source also compiled against clean `main` headers, and its four kernel tests passed.
- Tests, benchmarks, reports, and artifacts remain local and are not committed.

Measurements cover one image/model on this CPU with unlocked clocks; speed and numerical equivalence beyond the tested cases are not established.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -553,6 +553,7 @@ cc_library(
     },
     textual_hdrs = [
         "gemma/gemma-inl.h",
+        "gemma/vit_attention-inl.h",
     ],
     deps = [
         ":basics",
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,6 +96,7 @@ set(SOURCES
   gemma/tokenizer.h
   gemma/vit.cc
   gemma/vit.h
+  gemma/vit_attention-inl.h
   gemma/weights.cc
   gemma/weights.h
   io/blob_store.cc
--- a/gemma/vit.cc
+++ b/gemma/vit.cc
@@ -42,6 +42,7 @@
 #include "hwy/highway.h"
 // After highway.h
 #include "gemma/gemma-inl.h"
+#include "gemma/vit_attention-inl.h"
 #include "ops/ops-inl.h"
 
 HWY_BEFORE_NAMESPACE();
@@ -68,7 +69,7 @@ class VitAttention {
                layer_.vit.qkv_einsum_b.PackedScale1(), env_, qkv);
   }
 
-  // TODO(philculliton): transition fully to MatMul.
+  // QK uses the general matrix kernel; PV keeps FP32 values/probabilities.
   HWY_NOINLINE void DotSoftmaxWeightedSumMatrix() {
     const size_t qkv_dim = layer_config_.qkv_dim;
     const size_t heads = layer_config_.heads;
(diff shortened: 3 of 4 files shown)


Mikyx-1:
Closing: I mistakenly benchmarked this optimization against main instead of dev. The reported 1.97× speedup does not apply to dev, which already uses a different FlashAttention implementation.

jan-wassenberg:
:) Sorry about that. Yes, main is quite a ways behind, but it would indeed have been a nice optimization prior to our Flash attention.

PR #1020, opened 2026-09-06
Reduce local KV memory in BF16 Flash attention

### LINKED ISSUE #1016: Reduce KV memory for local-attention layers

Mikyx-1:
Closes #1016. Supersedes #1017, which was based on `main`.

The default Flash path on `dev` reserves full-context BF16 KV/K/V storage for local layers and also allocates compact tiled buffers it does not use. This change gives each owning Gemma layer separate BF16 buffers: local rings retain the attention window plus batch and SIMD padding, while global layers retain the configured context. The default Flash path no longer allocates unused tiled buffers.

Shared layers alias their source buffers. Ring growth preserves live flat and transposed history, snapshots remain independent, and cache recycling clears the owned buffers. API sessions and conversation bindings use the runtime-aware constructor. The legacy constructor and tiled, DeepSeek, and encoder/decoder layouts remain supported.

### Validation

Against `upstream/dev` at `1658f88`, all 13 A/B pairs on Gemma 3 270M and 1B produced identical decoded bytes and token IDs: 1,664 generated tokens per build. Runs used a shared baseline-calibrated matrix plan, six pinned workers on an i5-12400F, context 8,192, prefill batch 256, and 128 greedy output tokens.

Medians from three alternating pairs with a 3,367-token prompt:

| Model | Cache capacity, MiB | Peak RSS, MiB | Prefill, tokens/s | Decode, tokens/s |
|---|---:|---:|---:|---:|
| 270M | 503.0 → 73.7 | 638.0 → 559.2 | 1512.0 → 1499.5 | 46.89 → 46.57 |
| 1B | 705.5 → 101.4 | 1540.1 → 1430.7 | 259.4 → 259.6 | 14.14 → 14.10 |

The measured benefit is lower memory use; these results do not demonstrate a speedup. A slower single short-prompt 270M sample did not reproduce in three confirmation pairs (median decode 49.29 → 49.21 tokens/s).

- Seven regression tests passed in Release and under AddressSanitizer with leak detection. Coverage includes capacities, context caps, shared aliases, wraparound, growth, transposed history, snapshots, clearing, and 18 exact full-cache/ring kernel comparisons on AVX2 and EMU128.
- ASan instrumented the changed cache/attention units and tests; supporting libraries were reused without instrumentation.
- CLI, API server, single and batch benchmarks built; conversation binding compile check passed. Bazel dependencies were updated, but Bazel was not run.
- End-to-end model coverage is 270M and 1B; Gemma 4 shared allocation is tested by configuration. Vision and continuous multi-query inference were not benchmarked.

The commit contains production changes, regression tests, and their build wiring. Benchmark harnesses, reports, and generated artifacts are excluded.

### DIFF
--- a/gemma/attention.cc
+++ b/gemma/attention.cc
@@ -60,7 +60,7 @@ void TransposeKVCacheRow(const KV_t* HWY_RESTRICT kv, KV_t* HWY_RESTRICT k,
   // This is inefficient, as the writes are scattered over cache lines, but it
   // is a tiny fraction of the overall computation, and it is linear in the
   // token length.
-  const size_t kFloatsPerTile = 2 * FloatsPerVector();
+  const size_t kFloatsPerTile = 2 * hn::Lanes(hn::ScalableTag<float>());
   const size_t kRoundedQkvDim = hwy::RoundUpTo(qkv_dim, kMaxBF16PerVector);
   for (size_t i = 0; i < qkv_dim; i += 2) {
     k[i * kFloatsPerTile] = kv[i];
@@ -94,7 +94,7 @@ void TransposeKVCacheRow(const KV_t* HWY_RESTRICT kv, KV_t* HWY_RESTRICT k,
 
 void TransposeKVCacheRow_KEqV(const KV_t* HWY_RESTRICT kv, KV_t* HWY_RESTRICT k,
                               KV_t* HWY_RESTRICT v, size_t qkv_dim) {
-  const size_t kFloatsPerTile = 2 * FloatsPerVector();
+  const size_t kFloatsPerTile = 2 * hn::Lanes(hn::ScalableTag<float>());
   const size_t kRoundedQkvDim = hwy::RoundUpTo(qkv_dim, kMaxBF16PerVector);
   for (size_t i = 0; i < qkv_dim; i += 2) {
     k[i * kFloatsPerTile] = kv[i];
@@ -130,7 +130,7 @@ void TransposeKVCacheRow_KEqV(const KV_t* HWY_RESTRICT kv, KV_t* HWY_RESTRICT k,
 // positions.
 void TransposeOOBKVCacheRow(KV_t* HWY_RESTRICT k, KV_t* HWY_RESTRICT v,
                             size_t qkv_dim) {
-  const size_t kFloatsPerTile = 2 * FloatsPerVector();
+  const size_t kFloatsPerTile = 2 * hn::Lanes(hn::ScalableTag<float>());
   const size_t kRoundedQkvDim = hwy::RoundUpTo(qkv_dim, kMaxBF16PerVector);
   for (size_t i = 0; i < kRoundedQkvDim; i += 2) {
     k[i * kFloatsPerTile] = hwy::ConvertScalarTo<KV_t>(0.0f);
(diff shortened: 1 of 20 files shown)


gemma/attention.cc
```diff
@@ -219,26 +222,31 @@ static HWY_INLINE void ComputeQKV(size_t num_tokens, const size_t layer_idx,
...
+    MaybeReshapeCache(cols, cache.HasLayerCaches() ? cache.LayerV(kv_layer_idx)
+                                                   : view.v_cache);
   }
   const size_t kFloatsPerVector = FloatsPerVector();
-  const size_t kRoundedTokens =
-      hwy::RoundUpTo(num_tokens, 2 * kFloatsPerVector);
+  size_t kRoundedTokens = 0;
+  for (size_t qi = 0; qi < qbatch.Size(); ++qi) {
+    kRoundedTokens = HWY_MAX(
+        kRoundedTokens,
```
jan-wassenberg: This is not constexpr, please use normal variable naming (snake_case).

gemma/kv_cache.cc
```diff
@@ -265,6 +265,39 @@ KVCache::KVCache(const ModelConfig& config, const InferenceArgs& inference_args,
...
 
+  // Dense Gemma Flash consumes only these buffers. Avoid allocating both
+  // full-context legacy matrices and unused compact tiled matrices.
+  const bool dense_gemma =
+      !config.is_encoder_decoder && config.num_mtp_layers == 0 &&
+      std::all_of(kv_layer_configs.begin(), kv_layer_configs.end(),
+                  [](const LayerConfig& layer) {
+                    return layer.type == LayerAttentionType::kGemma;
+                  });
+  if (dense_gemma && runtime_config.attention_impl == AttentionImpl::kFlash) {
```
jan-wassenberg: RuntimeConfig is the older/deprecated mechanism. Prefer to pass attention_impl directly, and ideally later we'll move it to InferenceArgs; but definitely let's avoid further dependency on RuntimeConfig.

jan-wassenberg (commented):
Thank you! It seems like this would make a big difference for longer sequences.

But unfortunately, as-is, this is adding too much tech debt :/ We'd have a third layout: kv/k/v, compact, per-layer - semi-invisible, based on whether we use the 3-arg ctor or not. And the abstraction leaks, user code has to remember to check `ring`.

Let's think about alternatives. Can we use the compact everywhere in flash attention?

Mikyx-1:
Hi @jan-wassenberg, thank you for the helpful review! I've pushed 8b8e56a to address your concerns.

Following your suggestion, Flash backends now share compact per-head cache ownership. The default BF16 path uses reusable batch-local projection scratch, removing the third persistent layout and caller-side `ring` checks. Cache construction takes `AttentionImpl` directly without depending on `RuntimeConfig`, and the variable naming is fixed.

Validation: 51 relevant CTest cases and 31 focused AddressSanitizer cases passed. Six comparisons against a fresh build of the original implementation matched all 768 generated tokens, with roughly half the allocated KV storage and essentially unchanged throughput on the tested 270M and 1B cases.

Whenever you have a chance, would you mind taking another look? Thanks again!

gemma/gemma.cc
jan-wassenberg, line 95: Please revert :)

gemma/kv_cache.h
jan-wassenberg, line 43: Please also revert this and any other inadvertent changes :)

gemma/gemma.cc
Mikyx-1: Sorry about this, @jan-wassenberg. I made a mistake while resolving the merge conflicts and accidentally dropped upstream changes. Fixed in 47dbf1a: restored IsTiledAttention and both ZeroInit APIs, and removed the other inadvertent edits I found. Thanks for catching this!

PR #1030, opened 2026-09-12
Reduce default-flash RAM by reusing BF16 projection scratch

### LINKED ISSUE #1029: Default flash attention retains redundant BF16 KV projection history
## Summary

Default flash attention retains two BF16 representations of KV history.
The sequence-major projection buffer remains allocated and populated after
its keys and values have been transposed into the buffers attention reads.
This increases resident memory as context length grows.

## Affected path

- Backend: default `--attention_impl flash`.
- Confirmed with Gemma 3 270M, 1B, and 4B text inference.
- Measured baseline: `ffc1abc05abdf11d875d25d36c8859553ddf2641`.
- The retained-history path is also present on current `dev` (`b68def1`).
- T5 and DeepSeek use their legacy buffers differently.

## What happens

1. `ComputeQKV` writes BF16 projections into sequence-major `kv_cache`.
2. It applies normalization and positional encoding with BF16 rounding.
3. It transposes the results into `k_cache` and `v_cache`.
4. `FlashAttention` consumes those transposed buffers.
5. The original projection buffer still retains every layer and position.

References:
- [Projection and transpose path](https://github.com/google/gemma.cpp/blob/ffc1abc05abdf11d875d25d36c8859553ddf2641/gemma/attention.cc#L208-L322).
- [Full-sequence cache allocations](https://github.com/google/gemma.cpp/blob/ffc1abc05abdf11d875d25d36c8859553ddf2641/gemma/kv_cache.cc#L275-L297).

## Reproduction and observed cost

Run default-flash Gemma 3 270M with a 32,736-token text prompt,
sequence capacity 32,768, prefill batch 4,096, and 16 decode tokens.
Inspect the allocated cache extents and peak process RSS during inference.

The projection buffer occupies 578 MiB including row padding.
The transposed K/V buffers occupy another 576 MiB.
Measured peak process RSS is 1,778.59 MiB on this workload.
This is active retained data, not merely unused virtual address space.

Environment: Linux, Intel i5-12400F, six pinned threads,
Release AVX2/Haswell build, no oneDNN, and approximately 15.5 GiB RAM.

## Expected behavior

Projection intermediates should not retain a second complete KV history
after attention's persistent representation has been produced.
Required context, existing BF16 rounding, and model behavior must be preserved.

Mikyx-1:
Default flash attention retains a full sequence-major projection history
after transposing it into the K/V buffers consumed by attention.
This change reuses BF16 projection scratch and reduces peak process RAM
while preserving the projection rounding and attention computation.

Fixes #1029

## Implementation

- Reuse one layer of batch-sized BF16 projection scratch per query.
- Size it for the widest layer; grow it for larger batches or layer widths.
- Convert older-constructor caches when GemmaAttention first uses them.
- Preserve sequence capacity and the transposed K/V history.
- Copy the actual K/V buffers when cloning the new scratch layout.
- Skip unused compact-buffer allocations for the default-flash runtime path.
- Preserve normalization, RoPE positions, and both BF16 rounding boundaries.

## Measured RAM effect

Peak RSS includes weights, KV caches, and inference working buffers.
These percentages describe the whole process, not model-weight compression.

| Gemma 3 model | Prompt tokens | Before | After | Reduction |
|---|---:|---:|---:|---:|
| 270M | 32,736 | 1,778.59 MiB | 1,205.35 MiB | 32.2% |
| 1B | 16,352 | 2,369.89 MiB | 1,957.65 MiB | 17.4% |
| 4B | 4,064 | 5,754.25 MiB | 5,216.35 MiB | 9.3% |

For 270M, the projection allocation falls from 578.00 to 4.25 MiB.
Combined projection plus K/V allocations fall from 1,154.00 to 580.25 MiB.
Dormant virtual allocations are not counted as physical RAM savings.

## Measured speed effect

| Model | Prefill seconds, before → after | Decode tokens/s, before → after |
|---|---:|---:|
| 270M | 39.4237 → 39.3970 | 37.4104 → 37.8506 |
| 1B | 68.7325 → 68.5894 | 12.1040 → 11.9374 |
| 4B | 74.3011 → 74.3155 | 2.9791 → 3.0151 |

Observed decode changes: +1.2% for 270M, -1.4% for 1B, +1.2% for 4B.
Prefill is essentially unchanged. The small measured speedups are tentative:
16-token decode intervals and few repeats do not establish statistical significance.

## Measurement setup

- Baseline `ffc1abc`; measured fix `caa2e13`; Linux / Intel i5-12400F.
- Release AVX2/Haswell, six pinned threads, no oneDNN or spinning.
- 270M: capacity 32,768, batch 4,096, three stock-autotuning repetitions.
- 1B: capacity 16,384, batch 4,096; 4B: capacity 4,096, batch 1,024.
- 1B/4B: two repeats with fixed matmul candidates and forced continuation.
- 4B uses the existing padded checkpoint with `--map 1` in both versions.
- Timings are medians; all runs are serial with a RAM/swap watchdog.
- Model snapshots show no swap; system swap grew 162.25 MiB in the 4B baseline.

## Validation and scope

- Five focused cache/attention tests passed on AVX2.
- Coverage includes two queries, both constructor paths, heterogeneous widths,
  scratch growth, retained sequence capacity, cache copying, and T5 storage.
- Fixed-matmul verification: all 32 full-logit hashes match for each model.
- Each version also matches itself across two repetitions.
- That covers 8,388,608 logit values per model under identical continuation.
- T5, DeepSeek, and alternative tiled-backend inference storage are unchanged.
- No quantization, local-window rings, context truncation, or weight changes.
- The existing tiled test could not compile due to missing mock/span symbols.
- Only three implementation files and two test files are included.
- Benchmark artifacts, diagnostic harnesses, and reports are not committed.

### DIFF
--- a/gemma/attention.cc
+++ b/gemma/attention.cc
@@ -197,7 +197,6 @@ static HWY_INLINE void ComputeQKV(size_t num_tokens, const size_t layer_idx,
                                   ? static_cast<size_t>(layer_config.kv_share_layer_idx)
                                   : layer_idx;
   const bool skip_kv = (layer_config.kv_share_layer_idx >= 0) || (flags & kSkipKV);
-  const size_t cache_layer_size = activations.config.layer_configs[kv_layer_idx].CacheLayerSize();
 
   // The original qkv_einsum_w has shape [(heads + kv_heads * 2), qkv_dim,
   // model_dim], which we reshaped to (heads + kv_heads * 2) * qkv_dim rows.
@@ -205,9 +204,14 @@ static HWY_INLINE void ComputeQKV(size_t num_tokens, const size_t layer_idx,
              /*add=*/nullptr, env, activations.q);
 
   if (skip_kv) return;
-  // Set up MatMul row pointers for writing to KV, which consists of
-  // `kv_heads` pairs of (k, v) vectors. This safely handles wraparound
-  // because rows are computed modulo seq_len.
+  // Each query reuses one layer of BF16 projections for the current batch.
+  // Grow for callers that override the configured prefill batch size.
+  for (size_t qi = 0; qi < qbatch.Size(); ++qi) {
+    auto& kv = qbatch.KV(qi);
+    kv.cache->EnsureProjectionRows(num_tokens, layer.qkv_einsum_w2.Rows());
+    kv.kv_cache = kv.cache->kv_cache;
+  }
+  // Set up MatMul row pointers for `kv_heads` pairs of (k, v) vectors.
   MatPtrT<KV_t> kv_rows("kv", Extents2D(activations.pre_att_rms_out.Rows(),
                                         layer.qkv_einsum_w2.Rows()));
   for (size_t interleaved_idx = 0; interleaved_idx < num_interleaved;
@@ -219,12 +223,8 @@ static HWY_INLINE void ComputeQKV(size_t num_tokens, const size_t layer_idx,
     // --seq_len must be large enough to avoid wraparound.
     HWY_DASSERT(cache_pos < activations.SeqLen());
 
-    const size_t layer_offset = qbatch.KV(qi).cache->layer_flat_offsets.empty()
-        ? kv_layer_idx * cache_layer_size
-        : qbatch.KV(qi).cache->layer_flat_offsets[kv_layer_idx];
-
     env.row_ptrs[0][interleaved_idx] = reinterpret_cast<uint8_t*>(
-        qbatch.KV(qi).kv_cache.Row(cache_pos) + layer_offset);
+        qbatch.KV(qi).kv_cache.Row(token_idx));
   }
   kv_rows.AttachRowPtrs(env.row_ptrs[0].get());
   CallMatMul(activations.pre_att_rms_out, layer.qkv_einsum_w2,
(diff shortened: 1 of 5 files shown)


jan-wassenberg (approved):
Thank you, a single layer at a time sounds sensible and is a nice reduction.
I will attempt to merge this if the internal usage is unaffected.

Note that medium term, we hope to move from flash to tiled attention, which has different memory usage.

jan-wassenberg:
FYI internal tests failed but seemingly only because of an unused variable warning/error. I am patching the code.

jan-wassenberg:
Unfortunately gemma_test is also failing, with tensor dimension mismatch :/ Please add this to the test suite.

Abort at mat.cc:31: Assert to.Rows() == from.Rows() && to.Cols() == from.Cols(): 
hwy::Abort() in gcpp::CopyMat() during PrefillTBatch().

Mikyx-1:
• Hi  @jan-wassenberg, thanks for the feedback. Could you share a little more information about the CopyMat() failure?

  I tested gemma_test on the PR head and with the PR applied to the latest dev, using Gemma 3 270M and 1B, but could not reproduce the abort. In the public code, PrefillTBatch() also does not appear to reach CopyMat().

  Could you provide:

  - The model and command-line flags used
  - The full stack trace

I will investigate and fix it.

jan-wassenberg:
Thanks for looking into it. The model being used is gemma3-270m-sfp-it.sbs.
I believe there are no other command line flags. Our stack trace had no further detail, but it seems to be this code:

```
      // Record kv_cache if this is the last generate token or EOS.
      bool is_last_generate_token =
          pos + 1 == generate_length_limit_[query_index] +
                         queries_[query_index].size() ||
          gemma_->Config().IsEOS(token);
      if (is_last_generate_token) {
        CopyMat(all_queries[query_index].kv_cache.kv_cache,
                kv_caches[query_index].kv_cache);
      }
```

I realized this is an internal-only test, sorry about that.
It might be enough to change its usage of KVCache to the 4-arg form. It is not great that behavior differs based on that :/ That's something that would to good to clean up, but best done by us because there are some other internal users.

Will try patching the test..

jan-wassenberg:
FYI additional patches were required: a ZeroInit member function plus updating the matmul chunking. Thanks for adding this :)

PR #1032, opened 2026-09-13
Allow mmap for blob-aligned checkpoints

### LINKED ISSUE #1031: Allow mmap for checkpoints with a partial final page
# Allow mmap for checkpoints with a partial final page

## 1. Problem

`ChooseMode` disables mmap when the checkpoint size is not divisible by the system page size.

This restriction is unnecessary: whole-file mmap requires a page-aligned offset, but its length may end within a page.

Affected checkpoints therefore fall back to eagerly copying all weights into private process memory. This also loads unused weights immediately and prevents clean weight pages from being reclaimed or shared as file-backed pages.

## 2. Solution

Remove the file-length alignment check from `ChooseMode`.

Keep the existing tensor bounds validation and fallback to ordinary reads when mmap actually fails.

Mikyx-1:
Fixes #1031.

Checkpoints padded to 256-byte blob boundaries currently fall back to reading when their sizes are not multiples of the system page size. Relax the loader guard to `kBlobAlign` so these checkpoints can use mmap while retaining the alignment requirement for kernels that read ahead.

- Share `kBlobAlign` between the blob store and weight loader, and correct the comment about page-aligned file lengths.
- Keep the read fallback for lengths that are not multiples of 256, including explicitly requested mapping.
- Test a real blob-aligned partial-page mapping, explicit and automatic mode selection, unaligned fallback, and BF16 conversion precedence.

## Validation

- `weights_test`: 6 tests passed.
- `blob_store_test`: 2 tests passed.
- `git diff --check` and formatting checks passed.

### DIFF
--- a/gemma/weights.cc
+++ b/gemma/weights.cc
@@ -676,16 +676,6 @@ WeightsPtrs::Mode weights_internal::ChooseMode(uint64_t file_bytes,
     map = Tristate::kFalse;
   }
 
-  // Disable mapping if not padded to the base page size.
-  if (file_bytes % allocator.BasePageBytes() != 0) {
-    if (map == Tristate::kTrue) {  // Only complain if explicitly requested.
-      HWY_WARN("Unable to map non-padded file (%zu, %zu), reading instead.",
-               static_cast<size_t>(file_bytes >> 10),
-               allocator.BasePageBytes());
-    }
-    map = Tristate::kFalse;
-  }
-
   // Check for user override:
   if (to_bf16 == Tristate::kTrue && map == Tristate::kTrue) {
     HWY_WARN("Cannot have to_bf16 && map, to_bf16 takes precedence.");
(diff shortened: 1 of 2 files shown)


jan-wassenberg:
Thanks, I agree mmap and Windows will not fail with an unaligned length.
But we still have a potential issue: our kernels can read multiple vectors ahead. If files are unaligned, say length 4090, most HW/OSes will round up to 4 KiB, but this is insufficient for loading a full vector.
How about we relax the alignment requirement to kBlobAlign (256)?

Mikyx-1:
Thanks @jan-wassenberg! Updated the PR to require `kBlobAlign` (256-byte) alignment, as you suggested, and added regression coverage; all 8 loader and blob-store tests pass. Could you take another look when you have a chance?

jan-wassenberg (approved):
Nice, thanks for making the change!

PR #1038, opened 2026-09-21
Make the CMake test build work again

yupengtang:
`cmake -B build -DGEMMA_ENABLE_TESTS=ON` fails before it builds anything, and has for a while. This fixes the chain of reasons, on Ubuntu 22.04, GCC 11.4, CMake 3.22, an Intel i7-1355U.

**1. Test target names collide with Highway's.** Highway derives its test target names from the file basename exactly as we do, and it is fetched before this block runs, so three names are already taken:

```
CMake Error at CMakeLists.txt:241 (add_executable):
  add_executable cannot create target "image_test" because another target
  with the same name already exists.
```

The collisions are `compress_test`, `dot_test` and `image_test`. A side effect worth noting: those three of our tests were never built through CMake, since the name resolved to Highway's target of the same name.

**2. `GTest::Main` does not exist here.** It is the legacy name that `find_package(GTest)` defines. When googletest arrives through Highway's fetch, the target is `gtest_main`, so every test target failed with "links to target GTest::Main but the target was not found".

**3. `threading_test` includes `gmock/gmock.h`**, whose headers come with the `gmock` target rather than with gtest_main, so it failed with "fatal error: gmock/gmock.h: No such file or directory".

**4. `ops/dot_test.cc` had drifted out of C++17.** Unbuilt since the collision, it initialises `PackedSpan` with parentheses, which is C++20 aggregate initialisation, while this build sets `CMAKE_CXX_STANDARD 17`:

```
ops/dot_test.cc:944:24: error: no matching function for call to
  'gcpp::PackedSpan<float>::PackedSpan(float*&, const size_t&)'
```

It now uses the `MakeSpan` and `MakeConstSpan` helpers.

## After this change

Eleven test binaries build, and these run clean:

| test | result |
|---|---|
| basics_test | 5 tests pass |
| blob_store_test | 2 tests pass |
| distortion_test | 3 tests pass |
| fields_test | 8 tests pass |
| gemma_compress_test | 6 tests pass |
| gemma_image_test | 2 tests pass (run from the repo root, it reads `paligemma/testdata/image.ppm`) |
| nuq_test | builds, 0 tests on this target |
| ops_test | 30 tests pass |
| sfp_test | 14 tests pass |
| tensor_info_test | 1 test passes |

## Two things this does not fix

**`threading_test` and `gemma_dot_test` do not link**, because they use `hwy::Stats` and the pinned Highway (`2a16a50f`) does not compile `hwy/stats.cc` into any library. It is listed in Highway's `BUILD` but not in its `CMakeLists.txt` at that commit; Highway's current main lists it in both. So this needs a newer Highway pin, which felt like your call rather than something to slip into this change.

```
undefined reference to `hwy::Stats::ToString[abi:cxx11](int) const'
```

Compiling `hwy/stats.cc` into those two targets locally confirms the rest works: `threading_test` then passes all 8 tests.

**`gemma_dot_test` aborts on an accuracy expectation** once it links, which I did not touch since I cannot tell whether the bound is stale or whether this machine is genuinely outside it:

```
!!line 751, comp2 actual 2.586005E+00 not in [1.001000E+00, 1.400000E+00]
Abort at dot_test.cc:100: Assert false:
```

Happy to open a separate issue for either of those.

I used AI assistance (Claude Code) while working through this, mainly to bisect the configure and link failures and to draft this description. I ran every build and test listed above myself and reviewed each line of the change.


jan-wassenberg:
Thank you for helping maintain our tests :) We are only able to integrate changes to the dev branch - would you mind retargeting to that?

yupengtang:
Retargeted to `dev`, thanks.

Rebuilding it there was worth doing: `dev` already fixes most of what I had. It uses `find_package(GTest REQUIRED)`, so the `GTest::Main` alias is unnecessary; it sets C++20, so the `ops/dot_test.cc` change is unnecessary; and its newer Highway pin compiles `hwy/stats.cc`, so the `hwy::Stats` link error is gone. The `WORKING_DIRECTORY` you added to `gtest_discover_tests` also covers the relative-path problem I hit with `image_test`.

What is left is the target-name collision and the missing gmock link, 13 lines in `CMakeLists.txt`. On `dev` plus this change, `gemma_compress_test` (6), `gemma_image_test` (2), `threading_test` (6) and `ops_test` (40) all pass here.

One thing I did not touch: `gemma_dot_test` builds and links, then aborts on `comp2 actual 2.586005E+00 not in [1.001000E+00, 1.400000E+00]` on an Intel i7-1355U. I could not tell whether the bound is stale or whether this machine is genuinely outside it, so I left it for you. Happy to file it separately.

jan-wassenberg:
Thanks for updating. Unfortunately the number of commits is too large for our CLA scanner. Would you mind re-creating the PR so we can merge it?

